PR #10, opened 2020-08-10
Fix minor typos and rewording.

jdeisenberg:
Fixes minor typographical errors. Replace a reference to `foo` with a more meaningful name. Eliminate the word “simply“ (not only a word to avoid in technical writing, in the instance where I replaced it, it doesn’t fit).


chenglou:
Thanks!

PR #20, opened 2020-08-16
Correct JS output for char in overview

fhammerschmidt:
Compiling char to JS actually yields not a string but an int (the char code).


chenglou:
Ah yes, thanks

PR #23, opened 2020-08-19
Change confusing use of pure

yunti:
Remove the possibly confusing use of pure


chenglou:
I'll re-correct it. Thanks.

PR #25, opened 2020-08-19
A more positive strap line

yunti:
Very bikeshedding territory with this, but the strap line could definitely be more positive :) . This is just an example. 
Others could be `the javascript from the future`  or `A javascript with only the good parts, (and some extra great ones too)` .... etc...


chenglou:
I'll temporarily close this but it's worth reconsidering indeed

PR #39, opened 2020-08-27
Fix #38: Typo in let binding docs - rogue ';'

### LINKED ISSUE #38: Typo: Rogue ';' in 'let binding' docs
Currently there is a rogue ';' in the let binding section of the docs in the 'Design Decisions' section of the docs:

```re
if displayGreeting {
  let message = "Enjoying the docs so far?"
  Js.log(message)
}; // <~ here
```

My understanding is that this is no longer necessary to have a semicolon here, so it can be dropped. Please let me know if I am mistaken as I'm not 100% confident given the changeover to ReScript and optional semicolons.

bdunn313:
I believe that the semicolon in the design decisions section of the let binding doc is not necessary, so here is a PR to drop it.

If I'm wrong, let me know!


IwanKaramazow:
@bdunn313 indeed, thanks!

PR #60, opened 2020-09-22
New react docs (messed up branch)

ryyppy:
**Note**: These docs are not even remotely done and need a lot of stylistic refinement.


ryyppy:
This branch will be closed in favor of #96 since I messed up and made changes that are unrelated to the new React docs. In the future I need to be careful not doing that, even if it is tempting.

PR #61, opened 2020-09-23
part one on release 8.3

bobzhang:
@ryyppy does the current blogging support displaying two syntaxes?
Does the current engine supports `md` syntax as well? I wonder if I don't need jsx in md, could the engine do some work like renameing `md` into `mdx` automatically, I still use `md` syntax mostly because it has better tooling by default (preview mode without installing an extension)

### DIFF
--- a/_blogposts/2020-09-23-release-8-3.md.raw
+++ b/_blogposts/2020-09-23-release-8-3.md.raw
@@ -30,36 +30,15 @@ In this release, we make it optional to have `bs.` prefix it or not, this will m
 For example, the old externals for `readFileAsUtf8Sync` used to be written like this
 
 
-<CodeTab labels={["ReScript", "Reason", "ML"]}>
-```res
-@bs.val @bs.module("fs")
-external readFileAsUtf8Sync: (string, @bs.as("utf8") _) => string = "readFileSync"
-```
-```reason
-[@bs.val] [@bs.module "fs"]
-external readFileAsUtf8Sync: (string, [@bs.as "utf8"] _) => string =
-  "readFileSync";
-```
 ```ocaml
 external readFileAsUtf8Sync : string -> (_[@bs.as "utf8"]) -> string = "readFileSync" [@@bs.val] [@@bs.module "fs"]
 ```
-</CodeTab>
 
 Could be simplified as 
-<CodeTab labels={["ReScript", "Reason", "ML"]}>
-```res
-@val @module("fs") external readFileAsUtf8Sync: (string, @as("utf8") _) => string = "readFileSync"
-```
-```reason
-[@val] [@module "fs"]
-external readFileAsUtf8Sync: (string, [@as "utf8"] _) => string =
-  "readFileSync";
-```
 ```ocaml
 external readFileAsUtf8Sync : string -> (_[@as "utf8"]) -> string = "readFileSync" 
 [@@val] [@@module "fs"]
 ```
-</CodeTab>
 
 Note almost all previous attributes with `bs.xx` can be simplified as `xx`, except such two that 
 we don't provide abbreviations:
@@ -74,23 +53,11 @@ backward compatibility.
 
 If you use es6 module output, the default bindings will be compiled properly now:
 
-<CodeTab labels={["ReScript", "Reason", "ML"]}>
-```res
-@module("hello") external input: string => string = "default"
-
-let a = input("hello")
-```
-```reason
-[@module "hello"] external input: string => string = "default";
-
-let a = input("hello");
-```
 ```ocaml
 external input : string -> string = "default" [@@module "hello"]
 
 let a = input "hello"
 ```
-</CodeTab>
 
 Will now be compiled properly under es6 format as below:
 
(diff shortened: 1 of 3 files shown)


bobzhang:
Have no idea what's going on, need some help debugging next.js..

_blogposts/2020-09-23-release-8-3.mdx
```diff
@@ -0,0 +1,174 @@
...
+  },{
+    "module": "commonjs",
+    "suffix": ".cjs"
+  }],
+  
+```
+
+## More flexible filename support
+
+To have better integration with other [JS infrastructures](https://github.com/rescript-lang/rescript-compiler/issues/4624), for example, Next.js/React Native, we allow file names like `404.res`, `Button.Andriod.res` so that it can just be picked up by those tools
```
fhammerschmidt, line 120: Small typo: `Button.Android.res`

fhammerschmidt (changes requested):

yawaramin (approved):

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+npm i bs-platform@8.3.0
+```
+
+The changes are listed [here](https://github.com/rescript-lang/rescript-compiler/blob/master/Changes.md#83), this is a large release,  and we will go through some highlighted changes.
+
+
+
+## Lightweight FFI attributes without `bs.` prefix
+
+In this release, we make it optional to have `bs.` prefix it or not, this will make the FFI less verbose,
```
jdeisenberg, line 28: In this release, we make the `bs.` prefix optional. This will make the FFI less verbose.

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+
+For example, the old externals for `readFileAsUtf8Sync` used to be written like this
+
+
+```ocaml
+external readFileAsUtf8Sync : string -> (_[@bs.as "utf8"]) -> string = "readFileSync" [@@bs.val] [@@bs.module "fs"]
+```
+
+Could be simplified as
```
jdeisenberg, line 37: It can now be simplified as:

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+
+Could be simplified as 
+```ocaml
+external readFileAsUtf8Sync : string -> (_[@as "utf8"]) -> string = "readFileSync" 
+[@@val] [@@module "fs"]
+```
+
+Note almost all previous attributes with `bs.xx` can be simplified as `xx`, except such two that 
+we don't provide abbreviations:
+
```
jdeisenberg, line 45: Note that almost all previous attributes with `bs.xx` can be simplified as `xx` with the exception of the following two that don't have abbreviations:

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+- `bs.send.pipe` : this attribute was deprecated in favor of `bs.send`, you can still use the existing one for backward compatibility.
+
+- `bs.splice` : this attribute was deprecated in favor of `bs.variadic`, you can still use the existing one for 
+backward compatibility.
+
```
jdeisenberg, line 50: In lines 46 and 48, use a semicolon instead of a comma to be grammatically correct.

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+## More flexible filename support
+
+To have better integration with other [JS infrastructures](https://github.com/rescript-lang/rescript-compiler/issues/4624), for example, Next.js/React Native, we allow file names like `404.res`, `Button.Andriod.res` so that it can just be picked up by those tools
+
+
+
+## Better type based inference for pattern `let {a,b,c} = value`
+
+Previously for such piece of code:
```
jdeisenberg, line 93: Previously, for code like this:

_blogposts/2020-09-23-release-8-3.md.raw
```diff
@@ -0,0 +1,119 @@
...
+```
+
+You will get a type error
+
+```
+Error: Unbound record field x
+```
+
+However, since the compiler already knows the type of `u`, it is capable of looking up the label `x` properly, 
+in this release, we make such style work out of the box without the work around like adding a module prefix like `let {N.x} = ..`
```
jdeisenberg, line 113: ...properly. In this release, we make the original code style work out of the box without a work-around such as adding a module prefix like `let {N.x} = ..`

ryyppy:
> @ryyppy does the current blogging support displaying two syntaxes?

Discussed this feature offline (yes it is possible)


> Does the current engine supports `md` syntax as well? I wonder if I don't need jsx in md, could the engine do some work like renameing `md` into `mdx` automatically, I still use `md` syntax mostly because it has better tooling by default (preview mode without installing an extension)

I will create an issue to allow `md` files as well... but right now you are using JSX because of the multi syntax feature.

bobzhang:
@ryyppy do you have any ideal why the test fails?

jfrolich:
I was wondering why you would use both `bs.val` and `bs.module` at the same time in the example?

PR #66, opened 2020-09-29
Add active className in navigation menu

dmfilipenko:
When you are navigating over sub section in main menu, links are not highlights. 
Navigation
Drawbacks:
- Doesn't work on server-side. Server doesn't accept hash

### DIFF
--- a/bindings/Next.re
+++ b/bindings/Next.re
@@ -102,6 +102,7 @@ module Router = {
     route: string,
     asPath: string,
     events: Events.t,
+    pathname: string,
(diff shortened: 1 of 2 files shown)


ryyppy:
Hey @dmfilipenko. Sorry, I didn't have time to look at this for a while.

So this looks fine. We had plans on adding an intersection observer to our sidebar to update the highlighting depending on the section you are in. This would also solve the problem of highlighting the navigation item when loading the page initially.

I will merge this since this is a very simple change, so thanks for the effort.

PR #70, opened 2020-10-06
Small mention about disjunctive patterns 

AHBruns:
Disjunctive patterns are super valuable when you (rightly) don't want to use a wildcard, but have a lot of cases that match to the same thing. I think they should at least be mentioned in the docs even if they may not warrant a whole sub-section.

Because of the tiny size of this change, I just went ahead and opened a PR since it took about the same amount of work as opening an issue on the topic. I hope that's okay.

### DIFF
--- a/pages/docs/manual/latest/pattern-matching-destructuring.mdx
+++ b/pages/docs/manual/latest/pattern-matching-destructuring.mdx
@@ -583,6 +583,8 @@ Do not abuse the wildcard `_` too much. This prevents the compiler from giving y
 If you're about to use a wildcard when matching on a large but finite set of possibilities because you have many possibilities that return the same thing, consider reaching for disjunctive patterns instead. That is, you can replace:
 
 ```res example
+type T = A | B | C | D | E
+let v = C
 switch v {
 | A => handleTheCaseICareAbout()
 | B => handleTheRest()
@@ -595,6 +597,8 @@ switch v {
 with:
 
 ```res example
+type T = A | B | C | D | E
+let v = C
 switch v {
 | A => handleTheCaseICareAbout()
 | B
--- a/pages/docs/manual/latest/pattern-matching-destructuring.mdx
+++ b/pages/docs/manual/latest/pattern-matching-destructuring.mdx
@@ -580,6 +580,31 @@ Here are some advices.
 
 Do not abuse the wildcard `_` too much. This prevents the compiler from giving you better exhaustiveness check, which would be especially important after a refactoring where you add a new case to a variant. Try only using `_` against infinite possibilities, e.g. string, int, etc.
 
+If you're about to use a wildcard when matching on a large but finite set of possibilities because you have many possibilities that return the same thing, consider reaching for disjunctive patterns instead. That is, you can replace:
+
+```res example
+switch v {
+| A => handleTheCaseICareAbout()
+| B => handleTheRest()
+| C => handleTheRest()
+| D => handleTheRest()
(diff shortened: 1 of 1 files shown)


AHBruns:
Ah jeez. I'm sorry about trigger so many builds!

pages/docs/manual/latest/pattern-matching-destructuring.mdx
```diff
@@ -580,6 +580,35 @@ Here are some advices.
...
+let v = C
+switch v {
+| A => 1
+| B
+| C
+| D
+| E => 2
+}
+```
+This still requires a bit more work than a wildcard, but we promise it'll be worth it.
```
IwanKaramazow, line 610: `we promise it'll be worth it.` -> this is super vague. Can you illustrate/explain this?

IwanKaramazow: Also the rescript formatter allows the following kind of formatting:
```
switch v {
| A => 1
| B | C | D | E => 2
}
```

pages/docs/manual/latest/pattern-matching-destructuring.mdx
```diff
@@ -580,6 +580,35 @@ Here are some advices.
 
 Do not abuse the wildcard `_` too much. This prevents the compiler from giving you better exhaustiveness check, which would be especially important after a refactoring where you add a new case to a variant. Try only using `_` against infinite possibilities, e.g. string, int, etc.
 
+If you're about to use a wildcard when matching on a large but finite set of possibilities because you have many possibilities that return the same thing, consider reaching for disjunctive patterns instead. That is, you can replace:
```
IwanKaramazow, line 583: This sentence is very long. Can you divide it into shorter sentences?
Is it worth introducing the term `disjunctive patterns`?  What if we just talk about "writing out all cases explicitly"?

IwanKaramazow (changes requested):

pages/docs/manual/latest/pattern-matching-destructuring.mdx
AHBruns: I'll rewrite it to be more approachable. My only reason for using the term was to give readers a specific term that they can do more research on.

This brings up the larger idea of a glossary section which gives uncommon terms and brief definitions allowing people to grok concepts as well as giving them specific keywords with which to do independent research. This probably deserves an issue and longer discussion though.

AHBruns:
I refactored the intro to be simpler, and only mentioned the term disjunctive patterns as a brief aside. I also cleaned up the code examples and added a note about the refactoring advantages (mainly that you can have multiple disjunctive pattern sets unlike wildcards).

This is getting a bit large. So, it might be reasonable to split it into its own sub-section, but I'll leave that decision to you, of course.

ryyppy:
I like it! Maybe it would make sense to put this content further up in the the "Exhaustiveness Check" Section with a sub header like "Match multiple cases to one expression" or similar? That way it would not get lost in the Tips / Tricks section

AHBruns:
Moved it to its own sub-section. LMK if you'd like any other tweaks.

chenglou:
Hey @AHBruns. Thanks for the effort; this is indeed valuable to document more explicitly. I'll go ahead and merge it, but I'll rewrite it a bit.

chenglou:
Done. See my comments at https://github.com/reason-association/rescript-lang.org/commit/db82e572f2942a7e6c0d011df9102277610e14e8.

Thanks again!

PR #73, opened 2020-10-08
Better JS output in overview.mdx

theqp:
Use String Interpolation in JavaScript output at String Interpolation


ryyppy:
Hey! This output is based on the actual compiler output, so the way it is documented right now is already correct. See this [playground](https://rescript-lang.org/try?code=DYUwLgBAtiDOsEMDmIIF4ICIxzJgUKJAGYD2p6EABgBIjDAUAkA3jPMiAL5UDcQA) example.

PR #78, opened 2020-10-08
promise.mdx: replace deprecated pipe

theqp:
with the new version


theqp:
@ryyppy Thanks for the feedback. I couldn't find any formatter for ReScript, and your example is more different from the original version than mine.

ryyppy:
The [ReScript Playground](https://rescript-lang.org/try) has a "Format" button to pretty print your code.

The code looks different because the |> operator is actually an infix, and the -> operator is syntax. The latter doesn't introduce white space between the left- and righthand side values, so `values -> myFn` kinda looks off

theqp:
@ryyppy I have pretty printed the code snipet as suggested

ryyppy:
Thanks!

PR #82, opened 2020-10-17
add asterisk to Overview/if-else section

webmstk:


ryyppy:
Thanks for your quick fix. Could you also provide the same change to the `v8.0.0` docs? We still need changes over both versions until we made the full switch to the ReScript syntax.

webmstk:
Sure! Thanks for you work!

ryyppy:
Awesome! Thanks

PR #84, opened 2020-10-20
Fix Js.Dict2 mentioned in Js.Dict documentation

wingyplus:
The document is section for Js.Dict module but it's mentioned Js.Dict2
in function set section.

### DIFF
--- a/pages/docs/manual/latest/api/js/dict.mdx
+++ b/pages/docs/manual/latest/api/js/dict.mdx
@@ -60,7 +60,7 @@ Js.Dict.unsafeGet(ages, "Paul") // returns undefined
 let set: (t<'a>, key, 'a) => unit
 ```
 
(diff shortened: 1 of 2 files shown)


ryyppy:
I need to verify, but it might be that the most of the parameter lists of each Js.Dict function are wrong then.

Js.Dict2 has the major difference that the `dict` value goes in the first parameter position. In `Js.Dict` it's the last position (legacy pipe-last api).

wingyplus:
@ryyppy I'm not sure that `Js.Dict2` is simply to `Js.Dict` or it's a new module for ReScript. But I can make sure that `Js.Dict` is accepted `Js.Dict.t` as a first parameter like example below:

```
let () = {
  let dict = Js.Dict.empty()
  Js.Dict.set(dict, "hello", "world")
  Js.log(dict)
}
```

ryyppy:
Just noticed that `Js.Dict2` is actually not a thing. Thanks for your patience!

PR #87, opened 2020-10-28
Update jsx documentation

schicks:
The text implies that the formatter will prefer the punned version of code whenever possible. However, given the structure of the sentence that is the former option, not the latter. To avoid future ambiguity it seemed clearer to refer directly to "the punned syntax" rather than referencing positionally.


ryyppy:
Right, thanks for pointing that out. Could you also  add the same change to the equivalent `v8.0.0` document?

schicks:
Sorry that took so long, I was in the middle of moving into a new place.

ryyppy:
Awesome, thanks! Hope you had an easy time moving :)

PR #107, opened 2020-11-13
Update vim plugin url

amiralies:


ryyppy:
thanks! yeah we will adapt it later when we are getting closer to a proper release

PR #115, opened 2020-11-19
Fix incorrect example

Himself65:


ryyppy:
You have really good eyes. Thanks!

PR #126, opened 2020-11-30
Update overview.mdx

JoshuaSkootsky:
Take a look at this, I think the intended syntax here for Reason is `5 mod 3` as the equivalent of JS `5 % 3`.


ryyppy:
The latest docs are not in Reason, but in ReScript code.

In ReScript there are no infix operators yet, that's why it's  currently `mod(5,5)`

ryyppy:
I will close this now, because I think the docs are correct here. Feel free to reopen if necessary.

PR #137, opened 2020-12-15
Remove duplicate % from raw embed

Kingdutch:
It's the first example you see, but it doesn't actually work.
```
13 │ });
  14 │ 
  15 │ wsClient->addEventListener("connecting", %%raw(`console.log`));
  16 │ wsClient->addEventListener("connected", %%raw(`console.log`));
  17 │ wsClient->addEventListener("closed", %%raw(`console.log`));
  
  Did you forget a `)` here?
```

The singular form from other examples works as expected.

### DIFF
--- a/pages/docs/manual/latest/embed-raw-javascript.mdx
+++ b/pages/docs/manual/latest/embed-raw-javascript.mdx
@@ -13,7 +13,7 @@ First thing first. If you're ever stuck learning ReScript, remember that you can
 <CodeTab labels={["ReScript", "JS Output"]}>
 
 ```res example
-%%raw(`
+%raw(`
 // look ma, regular JavaScript!
 var message = "hello";
 function greet(m) {


ryyppy:
`%%raw` is used for module-level statements, so the example works just fine in the context of "declaring raw javascript on the toplevel":

[Playground Link](https://rescript-lang.org/try?code=KTBOEMHcAoAMCgD0iAEAbA9hg1igtuADQqgCmA5gK5rigoBS4AbuAMoDGoAlgA4AuAQngs6eUgGdx4cqRQBeFACIAFqTSZFAbngAzSgDt2fLhn0pyZUn2h4AlCgDe8FCnanxGNKQB0mcjdt4AF94WFsgA)

In your case, you need the `%raw()` decorator that attaches to an "expression or value" level, e.g.

```rescript
let foo = 1. +. %raw("1")
```

See [Playground Link](https://rescript-lang.org/try?code=DYUwLgBAZg9jEF4IEYB0EDU6CkAnAhgO4AUARMqQJRA)

Kingdutch:
I see! I did not get that distinction from the docs. That clears things up!

Then this PR isn't needed, thanks.

PR #138, opened 2020-12-17
Remove %obj() examples

kevanstannard:


ryyppy:
looking sharp. thanks!

PR #141, opened 2020-12-21
Add converters and helpers (bs.deriving)

ryyppy:


pages/docs/manual/latest/generate-converters-accessors.mdx
```diff
@@ -0,0 +1,565 @@
...
+
+</CodeTab>
+
+Variants constructors with payloads generate functions, payload-less constructors generate plain integers (the internal representation of variants).
+
+**Note**:
+- The generated accessors are lower-cased.
+- You can now use these helpers on the JavaScript side! But don't rely on their actual values please.
+- Doesn't work with polymorphic variants.
+
```
bobzhang, line 57: it does not work with poly var, the reason is that polyvar does not need it, the encoding of poly var is standardised

pages/docs/manual/latest/generate-converters-accessors.mdx
```diff
@@ -0,0 +1,565 @@
...
+let opaqueValue = fruitToJs(Apple)
+
+@bs.module("myJSFruits") external jsKiwi: abs_fruit = "iSwearThisIsAKiwi"
+let kiwi = fruitFromJs(jsKiwi)
+
+let error = fruitFromJs(100) /* nope, can't take a random int */
+```
+
+## Generate Converters for JS String Enums and Polymorphic Variants
+
```
bobzhang, line 277: The best practice: use jsConverter to convert between int and rescript types for regular variants. For string converter, it is not needed any more in ReScript

bobzhang (commented):
In general, this looks great!
I suggest we make a comment that user can use `deriving` directly

pages/docs/manual/latest/generate-converters-accessors.mdx
ryyppy: @bobzhang I believe we are still missing polyvariant docs to properly reflect that best practise, but I will add a version notice similar to the JS object converter

pages/docs/manual/latest/generate-converters-accessors.mdx
ryyppy: should i completely drop that line?

ryyppy:
> I suggest we make a comment that user can use deriving directly
Which minimal version allows `@deriving` instead of `@bs.deriving`?

bobzhang:
we can add a note that since 8.3, the bs. prefix is no longer needed

pages/docs/manual/latest/generate-converters-accessors.mdx
ryyppy: dropped it

pages/docs/manual/latest/generate-converters-accessors.mdx
ryyppy: added a note that since ReScript  >= 8.2 compiles polyvars to strings

ryyppy:
@bobzhang i added a notice in the beginning that starting from 8.3, ppl can drop the `bs.` prefix. When we lift the docs version number to 8.4 (after adding the rest of the missing docs), I'd completely remove the `bs.` from all examples and turn around the notice that `bs.*` is still supported but deprecated

PR #161, opened 2021-01-02
Sort index headers

kevanstannard:
Suggestion for https://github.com/reason-association/rescript-lang.org/issues/69


ryyppy:
Thanks. I like the detail with `t` as the first element.

PR #163, opened 2021-01-03
Add item.id to use as syntax lookup widget slug

kevanstannard:
Suggestion for https://github.com/reason-association/rescript-lang.org/issues/162

The `GithubSlugger` seems to convert non alphanumeric characters to hyphens.

For example `(.) => {}` is converted to `--`.

Which results the URL `/syntax-lookup#--`

What do you think about introducing an `id` field into the items? This would need to be unique for each item, but would provide a convenient field for converting to a slug.

Which would then enable URLs such as `/syntax-lookup#uncurried-function`


ryyppy:
Yeah that's better

Issue #164, opened 2021-01-03
Document how to call an uncurried function with a single `unit` argument


ryyppy:
This is a very common issue for learners (see [this SO question](https://stackoverflow.com/questions/65494134/how-do-you-call-an-uncurried-function-with-unit-in-rescript-reasonml/65548777#65548777)).

We should give some pointers in the [uncurried function section](https://rescript-lang.org/docs/manual/latest/function#uncurried-function) on how to call an uncurried function with a single unit argument:

E.g. this is a way to do it:

```rescript
let echo = (. a) => a
let _ = echo(. ignore())
```

ryyppy:
Fixed in #166

PR #166, opened 2021-01-04
Document calling an uncurried function with a single unit argument.

kevanstannard:
Suggestion for https://github.com/reason-association/rescript-lang.org/issues/164


ryyppy:
Thanks.

Github PRO tip: if you put in your description: `Fixes #164`, github will automatically close an issue the PR is linking to.

Issue #162, opened 2021-01-02
Syntax Widget: Technical Writing Tasks


ryyppy:
We've got the Syntax Lookup Widget (#111) in place now, so we need content writers to cover quick glance overviews over each topic.

This issue is here to collect a list of language constructs we should definitely cover. Feel free to propose more topics if something is missing. I will continuously update this issue for completeness reasons.

**Decorators:**

-  `@val`
-  `@module`
-  `@obj`
- `@scope`
- `@get`
-  `@get_index`
-  `@set`
- `@set_index`
- `@this`
- `@uncurry`
- `@send`
- `@string`
- `@int`
- `@unwrap`
- `@new`
- `@deriving`
- `@as`
- `@variadic`
- `@inline`
- `@meth`
- `@return`
- `@deriving`
- `@genType`
- `@unboxed`
- `@@warning`
- `@send.pipe` (deprecated)

**Operators:**
- `->`
- `|>`
- `++`
- `+`, `+.`
- `-`, `-.`
- `/`, `/.`
- `*`, `*.`
- `mod`
- `:>`
- `:=`

**Language Constructs:**

- `_` (placeholder, all scenarios like pipe, ignore value)
- `type`
- `let`
- `let \"\"` (bind to invalid identifier name)
- `if`
- `else`
- `external`
- `() => {}`
- `(.) => {}`
- `open`
- `include`
-  `module`
- `module type`
- `{.}`
- `{ "key": "value" }`
- `[ #value ]` (polyvar definition)
- `| Value` (variant, pattern matching)
- `switch`
- `import`
- `'a`
- `(Item: Comparable) => {}` (module functor)
- `exn`
- `ref`
- `exception`
- `| exception e` (exception pattern matching)
- `<MyComponent>` (JSX)
- `%%raw`
- `%raw`
- `%identity`
- j\`\`
- \`\`
- js\`\`

**Type / Value Specifics: **
- `Js.t`
- `type t`
- `ignore`

ryyppy:
Lots of things already shipped thanks to @kevanstannard's help!

We already got some Operators / Decorators documented, so I think it's already useful enough to merge it into master.

In case you want to open a new PR to add more content, please use `master` as your target branch in the future.

illusionalsagacity:
@ryyppy for the `:=` operator, is there a formal name for it? I have referred to it as the ref assignment operator in a branch, will publish a draft PR shortly

ryyppy:
[...]

PR #170, opened 2021-01-05
Syntax lookup dynamic include

kevanstannard:
Suggestion for dynamically importing the syntax files for #162

Usage is as follows:

[1] Write a new syntax mdx file.
[2] Run the `extract-syntax.js` script to generate a syntax index json file.
[3] The SyntaxLookupWidget will then read the index file and include on the page.

Just sharing what I have so far and see what you think.

image

Edit 1: I see the Vercel check has failed. I'm not sure what that problem might be.
Edit 2: Looks like adding extract-syntax to the update-index script fixed it.

### DIFF
--- a/misc_docs/syntax/controlflow_ifelse.mdx
+++ b/misc_docs/syntax/controlflow_ifelse.mdx
@@ -1,5 +1,9 @@
 ---
-test: "foo"
+id: "if-else"
+keywords: ["if", "else"]
+name: "if / else"
+summary: "This is the `if / else` control flow."
+category: "controlflow"
 ---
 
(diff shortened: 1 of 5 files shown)


ryyppy:
Right, so you solved one part, which is extracting the meta data back for the UI so it can do the search stuff.

My concern was more about the resulting bundle coming from the mdx component files (the help text for each syntax construct).

My initial idea was to use `getStaticProps` and try something like [mdx-remote](https://github.com/hashicorp/next-mdx-remote) to server side render the resulting markdown texts as plain strings, and later on insert them in the react tree with `__dangerouslySetInnerHtml` or so.

This change was definitely necessary anyways, so I will take it for now as is.

Thanks

PR #171, opened 2021-01-05
Document the @deprecated attribute. Fixes #168

### LINKED ISSUE #168: Document @deprecated decorator
Probably should be put in the [Attribute (Decorator) section](https://rescript-lang.org/docs/manual/latest/attribute#usage)

bdunn313:
Add a few examples for the @deprecated attribute. I've added it to the existing example, but alternatively it could show up as its on section on the attributes page. Fixes #168.

### DIFF
--- a/pages/docs/manual/latest/attribute.mdx
+++ b/pages/docs/manual/latest/attribute.mdx
@@ -45,6 +45,12 @@ type student = {
   age: int,
   @bs.as("aria-label") ariaLabel: string,
 }
+
+@deprecated
+let customDouble = foo => foo * 2
+
+@deprecated("Do the math yourself!")
+let customTriple = foo => foo * 3 
 ```
 ```js
 ```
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/attribute.mdx
```diff
@@ -45,6 +45,12 @@ type student = {
   age: int,
   @bs.as("aria-label") ariaLabel: string,
 }
+
+@deprecated
+let customDouble = foo => foo * 2
+
+@deprecated("Do the math yourself!")
```
ryyppy, line 52:
```suggestion
@deprecated("Use SomeOther.customTriple instead")
```
Would probably be a more practical use-case

bdunn313: Agreed, makes sense.

bdunn313: I needed to also update the explanation. After this builds, I believe it is ready.

ryyppy:
That's perfect. Thanks!

PR #172, opened 2021-01-05
Add note about externals and interface files

### LINKED ISSUE #159: Add note on External regarding interface files
When writing externals for JS code, it is rather easy to unintentionally introduce extra code by not using the same external in the interface (`.resi`) file. Therefore, whenever you are exposing an external, you need to make sure to copy the same external to your interface file to guarantee zero-cost bindings.

We previously had a note in the old docs that state

>Note: do also use externals and the [@bs.blabla] attributes in the interface files. Otherwise the inlining won't happen.

This is an important note, so it should be added back to the [External](https://rescript-lang.org/docs/manual/latest/external) page.

bdunn313:
Fixes #159. Ported note from original docs, but in new rescript syntax.

### DIFF
--- a/pages/docs/manual/latest/external.mdx
+++ b/pages/docs/manual/latest/external.mdx
@@ -78,6 +78,8 @@ However, if you want to more rigidly bind to the JavaScript library you want, ke
 
 Additionally, no extra ReScript-specific runtime is better for output readability.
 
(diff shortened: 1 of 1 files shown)


ryyppy:
Perfect, thanks!

PR #173, opened 2021-01-05
Add polyvar docs (latest)

### LINKED ISSUE #154: Document Polymorphic Variants
Very important construct that needs to be documented. Should be put into its own "Polymorphic Variant" section, right next to Variant I assume.

- Useful for interop scenarios, since they compile to strings (without payload)
- With payload, compile to a data structure that's friendly to type in tagged unions

ryyppy:
Fixes #154

### DIFF
--- a/pages/docs/manual/latest/polymorphic-variant.mdx
+++ b/pages/docs/manual/latest/polymorphic-variant.mdx
@@ -133,7 +133,7 @@ let other = [#GreeN]
 let all = Belt.Array.concat(colors, other)
 ```
 
-I'd get an error on the `concat` call, even thought he error actually stems from my typo in my `other` definition.
+I'd get an error on the `concat` call, even thought the error was actually caused by the typo in the value assignment of `other`.
 
 ## JavaScript Output
 
@@ -215,7 +215,7 @@ Since we were using poly variants, the JS Output is practically zero-cost and do
 
 ## Lower / Upper Bound Constraints
 
-There are a few different ways to define constraints on a poly variant type, such as `[>`, `[<` and `[]`. Some of them were briefly mentioned before, so in this section we will quickly explain what this syntax is about.
+There are a few different ways to define constraints on a poly variant type, such as `[>`, `[<` and `[`. Some of them were briefly mentioned before, so in this section we will quickly explain what this syntax is about.
 
 **Note:** We added this info for educational purposes. In most cases you will not want to use any of this stuff, since it makes your APIs pretty unreadable / hard to use.
 
@@ -237,7 +237,7 @@ In case you want to define a type that is extensible in polymorphic ways (or in 
 
 A lower bound defines the minimum set of constructors a poly variant type is aware of. It is also considered an "open poly variant type", because it doesn't restrict any additional values.
 
-Here is an example on how to make a minimum set of `basicBlueTones` extensible in a new `colors` type:
+Here is an example on how to make a minimum set of `basicBlueTones` extensible for a new `colors` type:
 
 ```res
 type basicBlueTone<'a> = [> #Blue | #DeepBlue | #Azuro ] as 'a
@@ -254,9 +254,9 @@ Here, the compiler will enforce the user to define `#Blue | #DeepBlue | #Azuro` 
 
 ### Upper Bound (`[<`)
 
-The upper bound works in the exact opposite way: The extending type may only use constructors that are stated in the lower bound constraing.
+The upper bound works in the exact opposite way: The extending type may only use constructors that are stated in the lower bound constraint.
 
-Same example:
+Here another example, but with red colors:
 
 ```res
 type validRed<'a> = [< #Fire | #Crimson | #Ash] as 'a
--- a/pages/docs/manual/latest/polymorphic-variant.mdx
+++ b/pages/docs/manual/latest/polymorphic-variant.mdx
@@ -10,7 +10,7 @@ Now that we know what [variant types](./variant) are, let's dive into a more spe
 
 First off, here are some key features:
 
-- Poly variants are structurally typed, which means they don't require any explicit type definition to be used as a value, and are not coupled to any specific module. The compiler will infer the type on demand, and compare poly variants by their value, instead of their type name (which is called nominal typing).
+- Poly variants are structurally typed, which means they don't require any explicit type definition to be used as a value, and are not coupled to any specific module. The compiler will infer the type on demand, and compare poly variants by their value, instead of their type name (which would be called nominal typing).
 - They allow easier JavaScript interop (compile to strings / objects with predictable `NAME` and `VAL` attribute) and don't need explicit runtime conversions, unlike common variants.
 - Due to their structural nature, poly variant types may cause tricky type checking errors when types don't match up.
 
@@ -81,7 +81,7 @@ var admin = "admin";
 
 </CodeTab>
 
-In rare cases (mostly for JS interop reasons), it's also possible to define invalid identifiers, such as hypens or numbers:
+In rare cases (mostly for JS interop reasons), it's also possible to define "invalid identifiers", such as hypens or numbers:
 
 <CodeTab labels={["ReScript", "JS Output"]}>
 
@@ -98,9 +98,11 @@ var oneA = "1a";
 
 </CodeTab>
 
+**Note:** The `\` character will be dropped in future syntax versions (way cleaner look, e.g. `#"1"`).
+
 ### Constructor Arguments
 
-This is equivalent to what we've already learned with common variants:
+This is equivalent to what we've [already learned](variant#constructor-arguments) with common variants:
 
 <CodeTab labels={["ReScript", "JS Output"]}>
 
--- a/pages/docs/manual/latest/polymorphic-variant.mdx
+++ b/pages/docs/manual/latest/polymorphic-variant.mdx
@@ -412,14 +412,13 @@ In case you want to define a type that is extensible in polymorphic ways (or in 
 
 A lower bound defines the minimum set of constructors a poly variant type is aware of. It is also considered an "open poly variant type", because it doesn't restrict any additional values.
 
-Here is an example on how to make a minimum set of `basicBlueTones` extensible for a new `colors` type:
+Here is an example on how to make a minimum set of `basicBlueTones` extensible for a new `color` type:
 
 ```res
 type basicBlueTone<'a> = [> #Blue | #DeepBlue | #LightBlue ] as 'a
 type color = basicBlueTone<[#Blue | #DeepBlue | #LightBlue | #Purple]>
 
 let color:  color = #Purple
-
 
 // This will fail due to missing minimum constructors:
 type notWorking = basicBlueTone<[#Purple]>
@@ -447,9 +446,9 @@ type notWorking = validRed<[#Purple]>
 
 One might think that polymorphic variants are fastly superior to common [variants](./variant). As always, it depends on the use case:
 
-- Variants allows better encapsulation for your APIs, since they require you to define a type definition that is coupled to a specific module.
+- Variants allow better encapsulation for your APIs, because they always come with a type definition that is coupled to a specific module.
 - Variants are conceptionally easier to understand, makes your code easy to refactor and provides better exhaustive pattern matching support
 - Variants usually deliver better type error messages, especially in recursive type definitions
 - Poly variants are useful for expressing strings in JS, and allow different type composition strategies. They can also be defined adhocly in your type definitions.
 
-In most scenarios, we'd recommend to use common variants over polymorphic variants, especially when you are writing plain ReScript code. In case you want to write zero-cost interop bindings, poly variants are a better option.
+In most scenarios, we'd recommend to use common variants over polymorphic variants, especially when you are writing plain ReScript code. In case you want to write zero-cost interop bindings or generate clean JS output, poly variants are oftentimes a better option.
--- a/pages/docs/manual/latest/polymorphic-variant.mdx
+++ b/pages/docs/manual/latest/polymorphic-variant.mdx
@@ -98,7 +98,7 @@ var oneA = "1a";
 
 </CodeTab>
 
-**Note:** The `\` character will be dropped in future syntax versions (way cleaner look, e.g. `#"1"`).
+**Note:** The `\` character will be dropped in future ReScript versions for a cleaner look (`#"1"`).
 
 ### Constructor Arguments
 
--- a/pages/docs/manual/latest/polymorphic-variant.mdx
+++ b/pages/docs/manual/latest/polymorphic-variant.mdx
@@ -10,28 +10,38 @@ Now that we know what [variant types](./variant) are, let's dive into a more spe
 
 First off, here are some key features:
 
-- Poly variants are **structurally typed** (in comparison to **nominally typed** variants). They can be used without an explicit type definition.
+- Poly variants are structurally typed, which means they don't require any explicit type definition to be used as a value, and are not coupled to any specific module. The compiler will infer the type on demand, and compare poly variants by their value, instead of their type name (which is called nominal typing).
 - They allow easier JavaScript interop (compile to strings / objects with predictable `NAME` and `VAL` attribute) and don't need explicit runtime conversions, unlike common variants.
-- Due their structural nature, they oftentimes cause tricky type checking errors when types don't match up, which makes them a more advanced feature.
+- Due to their structural nature, poly variant types may cause tricky type checking errors when types don't match up.
 
 ## Basics
 
-This is how you'd define a poly variant type with an exact set of constructors:
+Here is how you'd construct a poly variant value:
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res
+// Note how a poly variant starts with a hashtag (#)
+// We also don't need any explicit type definition
+let myColor = #Red
+```
+
+```js
+var myColor = "Red";
+```
+
+</CodeTab>
+
+This is how you'd define a closed poly variant type with an exact set of constructors:
 
 ```res
 // Note the surrounding square brackets, and # for constructors
 type color = [ #Red | #Green | #Blue ]
 ```
 
-Here is how you'd construct a poly variant value:
-
-```res
-// This doesn't actually need any color type definition
-// beforehand
-let myColor = #Red
-```
-
 We can also use poly variant types in annotations without an explicit type definition:
+
+<CodeTab labels={["ReScript", "JS Output"]}>
 
 ```res
 let render = (color: [#Red | #Green | #Blue]) => {
@@ -43,9 +53,21 @@ let render = (color: [#Red | #Green | #Blue]) => {
 let color: [#Red] = #Red
 ```
 
+```js
+function render(color) {
+  console.log("...");
+}
+
+var color = "Red";
+```
+
+</CodeTab>
+
 ### Constructor Names
 
 Poly variant constructor names are less restrictive than in common variants (e.g. they don't need to be capitalized):
+
+<CodeTab labels={["ReScript", "JS Output"]}>
 
 ```res
 type users = [ #admin | #moderator | #user ]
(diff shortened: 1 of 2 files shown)


pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+# Polymorphic Variant
+
+Now that we know what [variant types](./variant) are, let's dive into a more specific version, so called polymorphic variants (or poly variants).
+
+First off, here are some key features:
+
+- Poly variants are **structurally typed** (in comparison to **nominally typed** variants). They can be used without an explicit type definition.
+- They allow easier JavaScript interop (compile to strings / objects with predictable `NAME` and `VAL` attribute) and don't need explicit runtime conversions, unlike common variants.
+- Due their structural nature, they oftentimes cause tricky type checking errors when types don't match up, which makes them a more advanced feature.
```
jchavarri, line 15:
```suggestion
- Due to their structural nature, they oftentimes cause tricky type checking errors when types don't match up, which makes them a more advanced feature.
```

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+```res
+// Note the surrounding square brackets, and # for constructors
+type color = [ #Red | #Green | #Blue ]
+```
+
+Here is how you'd construct a poly variant value:
+
+```res
+// This doesn't actually need any color type definition
```
jchavarri, line 29: Maybe color could be something different than `Red` (like `Black`) in relation to use a value that is not in the type definition above.

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+// beforehand
+let myColor = #Red
+```
+
+We can also use poly variant types in annotations without an explicit type definition:
```
jchavarri, line 34:
```suggestion
We can also inline poly variant type definitions in annotations without an explicit type definition:
```

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+In rare cases (mostly for JS interop reasons), it's also possible to define invalid identifiers, such as hypens or numbers:
+
+```res
+type numbers = [#\"1" | #\"2"]
+let one = #\"1"
+```
+
+### Constructor Arguments
+
+This is equivalent to what we've already learned with common variants:
```
jchavarri, line 65: Maybe link to relevant section in common variants?

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+```
+
+As you can see in the example above, the type checker doesn't really care that `color` is not annotated as an `array<rgb>` type. As soon as it hits the first constraint (`Belt.Array.concat`), it will try to check if `colors` and `other` unify into one polymorphic type. If there's a mismatch, you will get an error on the `Belt.Array.concat` call.
+
+**That means that it is very easy to get confusing type errors on the wrong locations!**
+
+For instance, if I'd make a typo like this:
+
+```res
+let other = [#GreeN]
```
jchavarri, line 131:
```suggestion
let other = [#Green]
```

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+In case you want to define a type that is extensible in polymorphic ways (or in other words, subtyping allowed sets of constructors), you'll need to use the lower / upper bound syntax. 
+
+### Lower Bound (`[>`)
+
+A lower bound defines the minimum set of constructors a poly variant type is aware of. It is also considered an "open poly variant type", because it doesn't restrict any additional values.
+
+Here is an example on how to make a minimum set of `basicBlueTones` extensible for a new `colors` type:
+
+```res
+type basicBlueTone<'a> = [> #Blue | #DeepBlue | #Azuro ] as 'a
```
jchavarri, line 243: [Azure](https://en.wikipedia.org/wiki/Azure_(color))?

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+type myReds = validRed<[#Ash]>
+
+// This will fail due to unlisted constructor not defined by the lower bound
+type notWorking = validRed<[#Purple]>
+```
+
+## Tips & Tricks
+
+- In most scenarios, you should prefer common variants over polymorphic variants, since they offer better error messages and easier to spot errors in your program.
+- Polymorphic variants are pretty useful for doing zero-cost interop, e.g. when binding to JavaScript string enums, or to bind seemlessly between a tagged union type in TypeScript and ReScript
```
jchavarri, line 272: Another use case is when you need to "extend" an existing type with new values, which is briefly mentioned in PR but could be expanded: https://2ality.com/2018/01/polymorphic-variants-reasonml.html#extending-polymorphic-variants.

Extended variants also combines with the ability of pattern matching a whole set of values with an existing type: 

```ocaml
let extended_color_to_int : extended_color -> int = function
  | `RGBA (r,g,b,a) -> 256 + a + b * 6 + g * 36 + r * 216
  | #color as color -> color_to_int color
```

Not sure if you want to include all this though? :)

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: hehe, this is the actual trypo I tried to demonstrate: `For instance, if I'd make a typo like this:`

Maybe worth adding a comment on top

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: My supervisors told me to keep it short 😄 

The pattern matching part _might be interesting_ though.

Note to myself, example above looks like this in ReScript:

```rescript
let extendedColorToInt = (x: extendedColor): int => {
  switch x {
  | #RGBA(r, g, b, a) => 256 + a + b * 6 + g * 36 + r * 216
  | #...color as color => color_to_int
  }
}
```

pages/docs/manual/latest/polymorphic-variant.mdx
jchavarri: Oops, sorry, I missed the introductory line before. 😅

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+canonical: "/docs/manual/latest/polymorphic-variant"
+---
+
+# Polymorphic Variant
+
+Now that we know what [variant types](./variant) are, let's dive into a more specific version, so called polymorphic variants (or poly variants).
+
+First off, here are some key features:
+
+- Poly variants are **structurally typed** (in comparison to **nominally typed** variants). They can be used without an explicit type definition.
```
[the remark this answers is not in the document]
IwanKaramazow: Was also thinking about this. Do we have some explanations on rescript org we can link too?

pages/docs/manual/latest/polymorphic-variant.mdx
IwanKaramazow: Yes, same thought. Or switch the paragraphs. It's a bit strange that we first have a type declaration and then say we don't need it.

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+// Note the surrounding square brackets, and # for constructors
+type color = [ #Red | #Green | #Blue ]
+```
+
+Here is how you'd construct a poly variant value:
+
+```res
+// This doesn't actually need any color type definition
+// beforehand
+let myColor = #Red
```
IwanKaramazow, line 31: Can we add the js output?

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+---
+
+# Polymorphic Variant
+
+Now that we know what [variant types](./variant) are, let's dive into a more specific version, so called polymorphic variants (or poly variants).
+
+First off, here are some key features:
+
+- Poly variants are **structurally typed** (in comparison to **nominally typed** variants). They can be used without an explicit type definition.
+- They allow easier JavaScript interop (compile to strings / objects with predictable `NAME` and `VAL` attribute) and don't need explicit runtime conversions, unlike common variants.
```
IwanKaramazow, line 14: It would be great if we had some practical use cases/examples:
```rescript
@module("fs")
external openSync: (
  path,
  @bs.string
  [
    | @bs.as("r") #Read
    | @bs.as("r+") #Read_write
    | @bs.as("rs+") #Read_write_sync
    | @bs.as("w") #Write
    | @bs.as("wx") #Write_fail_if_exists
    | @bs.as("w+") #Write_read
    | @bs.as("wx+") #Write_read_fail_if_exists
    | @bs.as("a") #Append
    | @bs.as("ax") #Append_fail_if_exists
    | @bs.as("a+") #Append_read
    | @bs.as("ax+") #Append_read_fail_if_exists
  ],
) => unit = "openSync" 
```
```

File: pages/docs/manual/latest/polymorphic-variant.mdx [hunk 11]
```diff
@@ -0,0 +1,273 @@
...
+type users = [ #admin | #moderator | #user ]
+
+let admin = #admin
+```
+
+In rare cases (mostly for JS interop reasons), it's also possible to define invalid identifiers, such as hypens or numbers:
+
+```res
+type numbers = [#\"1" | #\"2"]
+let one = #\"1"
```
IwanKaramazow, line 60: Not sure if this is a good example, but `#\"r+"` could function as an example. As alternative to @bs.as
```rescript
@module("fs")
external openSync: (
  path,
  @bs.string
  [
    | @bs.as("r") #Read
    | @bs.as("r+") #Read_write
    | @bs.as("rs+") #Read_write_sync
    | @bs.as("w") #Write
    | @bs.as("wx") #Write_fail_if_exists
    | @bs.as("w+") #Write_read
    | @bs.as("wx+") #Write_read_fail_if_exists
    | @bs.as("a") #Append
    | @bs.as("ax") #Append_fail_if_exists
    | @bs.as("a+") #Append_read
    | @bs.as("ax+") #Append_read_fail_if_exists
  ],
) => unit = "openSync" 
```

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+  | #Instagram(string)
+  | #Facebook(string, int)
+]
+
+let acc: account = #Instagram("test")
+```
+
+### Annotations with Upper / Lower Bound Constraints
+
+There's also a way to define an "upper" and "lower" bound constraint for a poly variant type (that's why they are called _Polymorphic Variants_). Here is what it looks like in type annotations:
```
IwanKaramazow, line 79: What is the relation between "upper" and "lower" and polymorphic, I don't understand this very well.

pages/docs/manual/latest/polymorphic-variant.mdx
IwanKaramazow: Would an explanation of the `as 'a`  be needed here?

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+// This will fail due to missing minimum constructors:
+type notWorking = basicBlueTone<[#Purple]>
+```
+
+Here, the compiler will enforce the user to define `#Blue | #DeepBlue | #Azuro` as the minimum set of constructors when trying to extend `basicBlueTone<'a>`. 
+
+### Upper Bound (`[<`)
+
+The upper bound works in the exact opposite way: The extending type may only use constructors that are stated in the lower bound constraint.
```
IwanKaramazow, line 257:
```suggestion
The upper bound works in the exact opposite way: the extending type may only use constructors that are stated in the lower bound constraint.
```

pages/docs/manual/latest/polymorphic-variant.mdx
IwanKaramazow: Yes, I was also gonna suggest explaining the spread syntax. It's a special syntax thing.

pages/docs/manual/latest/polymorphic-variant.mdx
IwanKaramazow: Maybe as example next to the binding importance below.

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+type validRed<'a> = [< #Fire | #Crimson | #Ash] as 'a
+type myReds = validRed<[#Ash]>
+
+// This will fail due to unlisted constructor not defined by the lower bound
+type notWorking = validRed<[#Purple]>
+```
+
+## Tips & Tricks
+
+- In most scenarios, you should prefer common variants over polymorphic variants, since they offer better error messages and easier to spot errors in your program.
```
IwanKaramazow, line 271: Would it make sense to add a table with the differences between normal and poly variants summarized?

Should there be an example of when it's the wrong use case for poly variants?

IwanKaramazow (commented):
Good stuff!

pages/docs/manual/latest/polymorphic-variant.mdx
IwanKaramazow: Maybe as example in the binding importance section below.

pages/docs/manual/latest/polymorphic-variant.mdx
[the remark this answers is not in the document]
ryyppy: maybe it's better to explain the term "nominally typed" in the variants section, and later on link back to it? not sure where to put our definition of structurally typed.. down below we discuss the structural attributes anyways?

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: yeah it would, but not sure how to explain it in words without any concrete example (without being overly abstract)

pages/docs/manual/latest/polymorphic-variant.mdx
bobzhang: why do you need `@as` since you can use #\"r+" directly

pages/docs/manual/latest/polymorphic-variant.mdx
bobzhang: it is the combination of `structural + recursive types ` which makes the error message confusing

pages/docs/manual/latest/polymorphic-variant.mdx
bobzhang: in the backend, for polymorphic variant specifically, it is okay to use any names, and I think we should encourage people to use them directly without using `as`, that's part of the reason I proposed to make it a bit more native `#"hello"` instead of `#\"hello"`

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Don't worry about the upper / lower bound feature (aka polymorphism) just yet, since this is a very advanced topic that's often not really needed. For the sake of completeness, we mention a few details about it [later on](#lower--upper-bound-constraints).
+
+
+## Polymorphic Variants are Structurally Typed
+
+As we've already seen in the section above, poly variants are treated a little bit differently than common variants. Most notably, we don't need any explicit type definition to define a value.
+
+```res
```
bobzhang, line 101: You may mention one virtue of structural types, for package a and b who share the same structural types, they don't need depend on each other. But it is not the case for norminal types

pages/docs/manual/latest/polymorphic-variant.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+## Lower / Upper Bound Constraints
+
+There are a few different ways to define constraints on a poly variant type, such as `[>`, `[<` and `[`. Some of them were briefly mentioned before, so in this section we will quickly explain what this syntax is about.
+
+**Note:** We added this info for educational purposes. In most cases you will not want to use any of this stuff, since it makes your APIs pretty unreadable / hard to use.
+
+### Exact (`[`)
+
+This is the simplest poly variant definition, and also the most practical one. Like a common variant type, this one defines an exact set of constructors.
```
bobzhang, line 224: this is called closed types

pages/docs/manual/latest/polymorphic-variant.mdx
bobzhang: yeah, we should mention spreading syntax which is a nice sugar

pages/docs/manual/latest/polymorphic-variant.mdx
bobzhang, line 271: people have different preferences, in my experience, for recursive types, I would go with normal variants. But for simple data types, polymorphic variant is used from time to time

pages/docs/manual/latest/polymorphic-variant.mdx
austindd: The `as 'a` makes a lot more sense once we understand "type quantification," which I think is pretty foreign/arcane to a lot of people. But it might be helpful to allude to some of the **other** ways that types are quantified, and how the `as` keyword plays a role in determining how "type holes" are "filled."

ryyppy: We will have a dedicated section about polymorphism in an advanced section... i think i will go with a note that does some oversimplification for the case at hand, and later on point to more advanced resources when they are ready.

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: Switched paragraphs!

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: added an example that is more fitting and uses escape syntax

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: it's a feature that exists, so it needs to be documented

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: yeah that's confusing, removed that

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: I mentioned that somewhere in the variant / poly variant comparison in the end. Couldn't replicate any example where an error message gets completely out of hand, unless you are using very thorough libraries such as Tyxml

pages/docs/manual/latest/polymorphic-variant.mdx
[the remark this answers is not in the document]
ryyppy: I tried to explain it in a few words instead. I'd like to prevent wikipedia links, since they invite to go into an unnecessary rabbit hole

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: added in a small section

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: mentioned it in multiple occasions

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: removed the tips & tricks and added a "Variant vs Poly Variant" section instead with a few notes

ryyppy:
@bobzhang @IwanKaramazow @jchavarri thanks for your feedback! I addressed all your raised issues and actually added some more content that I forgot:

- Composing poly variant types (`type rgb = [ red | green | #Blue ]`)
- Pattern matching `| #...red`

etc etc

pages/docs/manual/latest/polymorphic-variant.mdx
ryyppy: i added a separate "Variants vs PolyVariants" section... but i refrained from bringing up examples because it would be too much

PR #183, opened 2021-01-13
Fix typo in docs for named arguments

phoenixeliot:
The order of arguments in the output was swapped compared to what's correct; that's all.

### DIFF
--- a/pages/docs/manual/latest/function.mdx
+++ b/pages/docs/manual/latest/function.mdx
@@ -124,7 +124,7 @@ You can provide the arguments in **any order**:
 addCoordinates(~y=6, ~x=5)
 ```
 ```js
-addCoordinates(6, 5);
+addCoordinates(5, 6);
 ```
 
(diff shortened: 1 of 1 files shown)


phoenixeliot:
Sharing my sanity checking just in case it helps anyone else sanity check:
https://rescript-lang.org/try
```
let addCoordinates = (~x, ~y) => {
  j`${Belt.Int.toString(x)}, ${Belt.Int.toString(y)}`
}

Js.log(addCoordinates(~x=5, ~y=6))
Js.log(addCoordinates(~y=6, ~x=5))
```
produces
```
function addCoordinates(x, y) {
  return String(x) + ", " + String(y);
}

console.log(addCoordinates(5, 6));
console.log(addCoordinates(5, 6));

exports.addCoordinates = addCoordinates;
```

bobzhang:
@phoenixeliot thanks!

PR #185, opened 2021-01-13
fix typo

dimitropoulos:
I noticed the other day but forgot to contribute: should be brackets here


ryyppy:
You are right. Thanks!

PR #187, opened 2021-01-13
Remove @@ infix examples from Js.Float docs

### LINKED ISSUE #184: What does \"@@"(Js.log, Js.Float.toString(param)) mean?

bdunn313:
Fixes #184

### DIFF
--- a/pages/docs/manual/latest/api/js/float.mdx
+++ b/pages/docs/manual/latest/api/js/float.mdx
@@ -65,12 +65,12 @@ Raises `RangeError` if `digits` is not in the range \[0, 20\] (inclusive).
 
 ```res example
 /* prints "7.71234e+1" */
-\"@@"(Js.log, Js.Float.toExponential(77.1234))
+Js.Float.toExponential(77.1234)->Js.log
 ```
 
 ```res example
 /* prints "7.7e+1" */
-\"@@"(Js.log, Js.Float.toExponential(77.))
+Js.Float.toExponential(77.)->Js.log
 ```
 
 ## toExponentialWithPrecision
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/api/js/float.mdx
```diff
@@ -65,12 +65,12 @@ Raises `RangeError` if `digits` is not in the range \[0, 20\] (inclusive).
 
 ```res example
 /* prints "7.71234e+1" */
-\"@@"(Js.log, Js.Float.toExponential(77.1234))
+77.1234->Js.Float._NaN.toExponential->Js.log
```
bdunn313, line 68: @ryyppy I'm frankly not sure what convention is here - I see it kind of go back-and-forth in the community. Should it be what I have here, or something like:

```suggestion
Js.Float._NaN.toExponential(77.1234)->Js.log
```

I can't recall if convention is to write the first function as above and then pipe to all subsequent functions, or to just start piping from the get-go. Let me know what the idiomatic call is here, if I have to change it, no big deal.

ryyppy: i think the latter form fn(arg)->Js.log makes more sense here. There is no general rule on the style though

ryyppy: ah and there seems to be some problem with the example (it doesn't compile). Use `npm run test` to compile the examples.

bdunn313: d'oh. That was a rookie mistake! 

I'll update all examples to match the fn(arg)->fn pattern, and _actually run tests before submitting_ this time. Thanks for the polite and patient nudge!

bdunn313: Good to go, closing this thread.

ryyppy:
perfect. thanks!

PR #192, opened 2021-01-15
Add syntax lookup pipe operator

kevanstannard:
Suggestion for #162

### DIFF
--- a/misc_docs/syntax/operators_pipe.mdx
+++ b/misc_docs/syntax/operators_pipe.mdx
@@ -6,7 +6,7 @@ summary: "This is the `pipe` operator."
 category: "operators"
 ---
 
-The `pipe` operator provides an convenient syntax for passing a value into a function. Typically it's used to pass a value into a function as its first argument. 
+The `->` operator provides an convenient syntax for passing a value into a function. Typically it's used to pass a value into a function as its first argument. 
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


misc_docs/syntax/operators_pipe.mdx
```diff
@@ -0,0 +1,45 @@
...
+  return "Hi " + name + ", you rolled a " + value.toString();
+}
+
+var message = dieRollMessage(Js_math.random_int(1, 6), "Marshall");
+```
+
+</CodeTab>
+
+Which produces a message such as `Hello Marshall, you rolled a 3`.
+
```
ryyppy, line 42:
~~~suggestion

You can also explicitly define the argument position of a piped value by using the `_` placeholder: 

<CodeTab labels={["ReScript", "JS Output"]}>

```res example
let logMsg = (user: string, datestr: string, msg: string): unit => {
  Js.log(`${user}|${datestr}|${msg}`)
}

let datestr = "01-01-2021"
let user = "admin"

// Here, we put the result of toUpperCase into the last position
// denoted with an _
Js.String2.toUpperCase("example message")->logMsg(user, datestr, _)
```

```js
function logMsg(user, datestr, msg) {
  console.log(user + "|" + datestr + "|" + msg);
  
}

var datestr = "01-01-2021";

var user = "admin";

logMsg(user, datestr, "example message".toUpperCase());
```

</CodeTab>
~~~

ryyppy (changes requested):
Proposed a few diffs. I think the `_` placeholder is also quite important to know

ryyppy:
Thanks again! 🚀

PR #193, opened 2021-01-15
Add syntax lookup triangle pipe operator

kevanstannard:
Suggestion for #162


misc_docs/syntax/operators_triangle_pipe.mdx
```diff
@@ -0,0 +1,47 @@
+---
+id: "triangle-pipe"
+keywords: [pipe", "triangle", "operator", "function", "argument"]
+name: "|>"
+summary: "This is the `triangle pipe` operator."
+category: "operators"
```
ryyppy, line 6: at some point we should probably introduce some `deprecated: true` field, to "grey out" labels in the overview maybe?

ryyppy:
Looking good!

PR #194, opened 2021-01-15
Installation in an existing project documentation

kevanstannard:
I was going through the installation instructions for [integrating ReScript into an existing project](https://rescript-lang.org/docs/manual/latest/installation#integrate-into-existing-js-project) today.

After installing `bs-platform` the guide indicates to follow the steps for a new project, which refers to cloning the ReScript template, etc.

I'm assuming that when integrating into an existing project then we would not follow the template clone steps? If that's the case, then I've drafted some ideas in this PR.

Any feedback appreciated.

Also, just confirming that we should use "JS" as the convention when referring to JavaScript?


pages/docs/manual/latest/installation.mdx
```diff
@@ -34,12 +34,62 @@ Alternatively, **to start a [ReasonReact](https://reasonml.github.io/reason-reac
 
 ## Integrate Into Existing JS Project
 
-You can install the toolchain locally to an existing JS project, through the familiar command:
+You can install ReScript locally into an existing JS project using the following steps.
```
ryyppy, line 37:
```suggestion
You can install ReScript locally into an existing JavaScript project using the following steps.
```

probably for this particular case, JavaScript is better.. but we use both versions

ryyppy (commented):
I think this is a good introduction, nothing to complain here.

kevanstannard:
Thanks @ryyppy, ready for re-review.

PR #197, opened 2021-01-16
Add syntax lookup mod operator

kevanstannard:
It looks like `mod` doesn't support infix syntax at the moment, but presumably it's fine to treat it as an operator for now? What do you think?


ryyppy:
Yes it is fine to treat it as an operator. There will be a dedicated doc for custom infix operators soon, but the implementation will follow a little later.
 
For now it looks fine

PR #200, opened 2021-01-17
Add "Import a JSON file content" to "Import From JavaScript" page

MoOx:
Could be handy for new comers or eternal newbies like me.

### DIFF
--- a/pages/docs/manual/latest/import-from-export-to-js.mdx
+++ b/pages/docs/manual/latest/import-from-export-to-js.mdx
@@ -70,6 +70,25 @@ console.log(Student.default);
 
 </CodeTab>
 
+### Import a JSON file content
+
+Use the `bs.module` [external](external.md):
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type dataType
+@bs.module external studentNames: dataType = "./students.json"
+Js.log(studentNames)
+```
+```js
+var StudentsJson = require("./students.json");
+var studentNames = StudentsJson;
+console.log(studentNames);
+```
+
+</CodeTab>
+
 ## Export To JavaScript
 
 As mentioned in ReScript's idiomatic [Import & Export](import-export.md), every let binding and module is exported by default to other ReScript modules. If you open up the compiled JS file, you'll see that these values can also directly be used by another _JavaScript_ file too.


pages/docs/manual/latest/import-from-export-to-js.mdx
```diff
@@ -70,6 +70,25 @@ console.log(Student.default);
...
 </CodeTab>
 
+### Import a JSON file content
+
+Use the `bs.module` [external](external.md):
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type dataType
```
kevanstannard, line 80: What do you think about making this a Json type `type dataType = Js.Json.t`?

pages/docs/manual/latest/import-from-export-to-js.mdx
```diff
@@ -70,6 +70,25 @@ console.log(Student.default);
 
 </CodeTab>
 
+### Import a JSON file content
+
+Use the `bs.module` [external](external.md):
```
kevanstannard, line 75: Very minor: Since the bs.module and link is mentioned earlier in the page, what do you think about replacing this with a simple sentence such as "ReScript also provides built in support for reading JSON files" (similar to the couple of previous examples).

ryyppy:
I think this particular example is too concrete. Import / Export talks about how to interop with different module kinds.
Maybe it would be better to put that in the Cheatsheet in the `JS Module Interop` section instead?

MoOx:
Yeah maybe. Also my example is incomplete as it's not the same thing that's need to be written depending on you `package-specs.module` value (`es6` or `commonjs`)
I wrote 

```res
@bs.module external studentNames: dataType = "./students.json"
```
That's for commonjs.

For `es6` you need to write

```res
@bs.module("./students.json") external studentNames: dataType = "default"
```

Anyway this definitely needs to be somewhere in the doc. I am doing ReScript/Reason for while now (not 100% of my weeks but still) and I always find my self to check doc for this to not have to think about it.

ryyppy:
Alriiight, I think we want to have this in, but the PR is a little outdated. Will copy over the changes and do some adaptions!

fhammerschmidt:
Done here: https://github.com/rescript-association/rescript-lang.org/commit/f5ec8978e5087e29116d02c8eecf0931bcbbb9c0

Issue #203, opened 2021-01-18
Array destructuring example in docs results in JS error after compilation


thepeoplesbourgeois:
I'm unsure of how to categorize this one. It definitely feels like a bug, given the documentation, but I'm almost certain it's a feature, given how strict the language seems as a whole.

The concerned section of documentation…

> Destructuring works with most built-in data structures:
>
> ```reason
> ...
>
> // Array
> let myArray = [1, 2, 3]
> let [item1, item2] = myArray // 1 assigned to `item1`, 2 assigned to `item2`
> 
> ...
> ```

… with the corresponding JS output, already declaring that it will fail:

> ```js
> var myArray = [1, 2, 3];
> if (myArray.length !== 2) {
>   throw {
>     RE_EXN_ID: "Match_failure",
>     _1: [
>       "playground.res",
>       14,
>       4
>     ],
>     Error: new Error()
>   };
> }
> var item1 = myArray[0];
> var item2 = myArray[1];
> ```

Indeed, plugging this output directly into console, unsurprisingly, returns the error:

```js
Uncaught Object { RE_EXN_ID: "Match_failure", _1: (3) […], Error: Error }
  <anonymous> debugger eval code:12
```

I installed `bs-platform` and cloned the `rescript-project-template` repo, in case there was some error handler in the larger BS platform that the [playground](https://rescript-lang.org/try) was abstracting away, but it was to no avail:

```sh
$ cat src/Demo.res
let myArr = [1,2,3]
let [one, two] = myArr

Js.log2(one, two)

$ node src/Demo.bs.js 

/Users/newb/workspace/javascript/rescript-tryout/src/Demo.bs.js:12
  throw {
  ^
{
  RE_EXN_ID: 'Match_failure',
  _1: [ 'Demo.res', 2, 4 ],
  Error: Error
      at Object.<anonymous> (/Users/newb/workspace/javascript/rescript-tryout/src/Demo.bs.js:19:16)
      at Module._compile (node:internal/modules/cjs/loader:1108:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
      at Module.load (node:internal/modules/cjs/loader:973:32)
      at Function.Module._load (node:internal/modules/cjs/loader:813:14)
      at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:76:12)
      at node:internal/main/run_main_module:17:47
}
```

[...]

Issue #198, opened 2021-01-16
Remove non exhaustive list binding from docs


amiralies:
From docs:
https://rescript-lang.org/docs/manual/latest/pattern-matching-destructuring

```rescript
// List
let myList = list{1, 2, 3}
let list{head, ...tail} = myList // 1 assigned to `head`, `list{2, 3}` assigned to tail
```
Compiler currently crashes in such case.
but aside that, since the the pattern fails for an empty list  I think we should remove this or warn the user about that.

kevanstannard:
Since the example is documenting pattern matching destruction for lists (and arrays above it) what do you think about updating the example rather than removing it?

For example:

```res
let myList = list{1, 2, 3}

let (head, tail) = switch myList {
| list{head, ...tail} => (Some(head), tail)
| list{} => (None, list{})
}
```

amiralies:
This is too complex in the context of destructuring IMO. I prefer adding a note explaining it can be fail and the pattern is not exhaustive.

kevanstannard:
Understood, I see your concern.

What about moving that content to a separate section "Destructuring Arrays and Lists" which includes examples like above.

Including a note explaining that list/array structuring can fail and that the pattern not being exhaustive is good information, but it doesn't help people new to ReScript deal with that problem. And I imagine those people are a primary target audience for the docs.

In other words, if we describe a problem then it would be good to also provide a guide to resolve it (either a link to another page, or describe the solution in the same place).

Also, maybe there is a simpler example we could use?

What do you think?

ryyppy:
This example should now be documented properly in master:

![image](https://user-images.githubusercontent.com/1121889/106566967-58a5dd00-6531-11eb-824d-8059b1f2a1e0.png)


I think we can close this issue?

amiralies:
Yup, Thank you.

PR #204, opened 2021-01-19
Update destructuring arrays and lists

kevanstannard:
Suggestion for #198 and #203 

There was some concern in one of the issues that these examples might be a bit complex for the destructuring section of the docs, and I can see the point there, and in that case maybe there is a better place to document this?

At minimum this provides some working examples.

Opening this PR for discussion. Happy to be guided by the team.

### DIFF
--- a/pages/docs/manual/latest/pattern-matching-destructuring.mdx
+++ b/pages/docs/manual/latest/pattern-matching-destructuring.mdx
@@ -44,57 +44,109 @@ let student1 = {name: "John", age: 10}
 let {name} = student1 // "John" assigned to `name`
 
 // Variant
-type result =
-  | Success(string)
+type result = Success(string)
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks! This got superseded but appreciate the PR. Here's some guidance:
- Generally speaking, we'd like to keep the examples related to the other sections, and have fewer overlapping examples.
- We shouldn't remove the top level matching, because then someone will file an issue to (re)introduce those examples or something. Instead, I'm keeping them and putting a warning in the related commit, which also serves as a transition to the next paragraph.
- Great of you to use real variable values and real variable names instead of `a` and `b` and such. Thanks for that.
- Your example uses Belt.List.length, which is heavily discouraged (it linearly goes through the list to count the number of items).
- The example should be in that section about matching on things. I've put it after matching on exceptions.

PR #218, opened 2021-02-04
Add deserializeUnsafe and serializeExn

### LINKED ISSUE #208: Add Js.Json.deserializeUnsafe and Js.Json.serializeExn to the docs
It looks like `Js.Json.deserializeUnsafe` and `Js.Json.serializeExn` are not currently in the docs.

https://rescript-lang.org/docs/manual/latest/api/js/json

Should they be added? Or perhaps intentionally omitted?

kevanstannard:
Fixes #208

As discussed, suggestion content for these functions.

### DIFF
--- a/pages/docs/manual/latest/api/js/json.mdx
+++ b/pages/docs/manual/latest/api/js/json.mdx
@@ -291,3 +291,39 @@ let stringifyAny: 'a => option<string>
 /* prints `["hello", "world"]` */
 Js.log(Js.Json.stringifyAny(["hello", "world"]))
 ```
+
+## deserializeUnsafe
+
+```res sig
+let deserializeUnsafe: string => 'a
+```
+
+Deserialises a JSON string to a type you specify.
+
+Raises a `Js.Exn.Error` exception if the string parsing fails.
+
+```res
+type person = {name: string}
+let p: person = Js.Json.deserializeUnsafe(`{ "name": "Albert" }`)
+```
+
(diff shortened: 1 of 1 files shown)


ryyppy:
Sorry for my slow responsiveness. The problem is that these functions are not serializing / deserializing JSON data to string and vice versa. They are serializing / deserializing records (that e.g. contain `option` value attributes) into a JSON compliant JSON object... really hard to explain, because as I said, I am not fully aware of the edge-cases and implications on more complex records.

I actually wrote a whole post about the original problem this api tried to solve [here](https://dev.to/ryyppy/reason-records-nextjs-undefined-and-getstaticprops-5d46).

Try it on a node script and see what `Js.log` prints.

ryyppy:
So the correct description should go along the line of: those functions are useful to quickly serialize / deserialize a record into a JSON compliant string... probably also works for other ReScript internal data structures, but not sure.

kevanstannard:
@ryyppy thanks for the feedback - copy updated.

Can you please confirm this is correct for `deserializeUnsafe`:

> Deserializes a JSON string created by `serializeExn()` into a ReScript record.

Feel free to suggest improvements.

kevanstannard:
Hi @ryyppy just following up on this one.

I was thinking about this one some more. Do you feel these functions are still worth documenting? Would you consider them experimental?

Happy to be guided by you.

kevanstannard:
👍 sounds good. Closing.

PR #223, opened 2021-02-15
Clarify comparison with js

glor:
Replace 'us' with ReScript as looks nicer and more intuitive and cannot be confused with the US which it did for me because it's rendered in caps on the website.


ryyppy:
thanks, like that better as well

PR #225, opened 2021-02-16
[docs/react] Fix keys and arrays example

### LINKED ISSUE #224: [docs/react] Incorrect example for item index as key
I tried to use the example for using item index as key on the new [rescript-react doc](https://rescript-lang.org/docs/react/latest/arrays-and-keys), but it doesn't work.

On the third example, the arguments for `Belt.Array.mapWithIndex` are reversed and the key attribute on the `li` tag should be converted to string.

marcagba:
Fixes #224

#### Changes
* Reverse parameters  of the `Belt.Array.mapWithIndex` call
* Convert index `i` to string
* Also highlight the first line of the example (as a newcomer to rescript you're likely to ignore that you need a different function to access index within a map call).

#### Before
Capture d’écran 2021-02-17 à 00 28 10

#### After
Capture d’écran 2021-02-17 à 00 27 59

### DIFF
--- a/pages/docs/react/latest/arrays-and-keys.mdx
+++ b/pages/docs/react/latest/arrays-and-keys.mdx
@@ -41,10 +41,10 @@ let items = Belt.Array.map(todos, todo => {
 
 If you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort:
 
-```res {2,3}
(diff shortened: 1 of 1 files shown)


ryyppy:
woops, thanks!

PR #229, opened 2021-02-20
Small docfix

vishnugopal:
The current parser seems to require braces `{` around the `React.string` call. Also updated the formatting to what is in my VS Code.

### DIFF
--- a/pages/docs/react/latest/rendering-elements.mdx
+++ b/pages/docs/react/latest/rendering-elements.mdx
@@ -44,8 +44,8 @@ Here is a full example rendering our application in our `root` div:
 // Dom access can actually fail. ResScript
 // is really explicit about handling edge cases!
 switch(ReactDOM.querySelector("#root")){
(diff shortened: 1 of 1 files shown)


vishnugopal:
Note that without this change, this is what I get:

```
>>>> Start compiling 
bsb: [3/3] src/Test-RescriptReact.cmj
FAILED: src/Test-RescriptReact.cmj

  We've found a bug for you!
  /Users/vishnugopal/Developer/Sandbox/rescript-react/src/Test.res:9:39-50
  
   7 │ // is really explicit about handling edge cases!
   8 │ switch ReactDOM.querySelector("#root") {
   9 │ | Some(root) => ReactDOM.render(<div> React.string "Hello Andrea" </di
       v>, root)
  10 │ | None => () // do nothing
  11 │ }
  
  This call is missing an argument of type string
```

ryyppy:
Thanks!

PR #230, opened 2021-02-20
Add documentation for rendering components to DOM

vishnugopal:
Because I found this confusing when starting out, especially if I needed to do an explicit import.


ryyppy:
@vishnugopal the [Rendering Elements](https://rescript-lang.org/docs/react/latest/rendering-elements#rendering-elements-to-the-dom) section is probably not specific enough right?

vishnugopal:
@ryyppy Yeah I found that it doesn't really cover how to render components. For me especially it was not obvious that files would be included automagically, and you could just reference components  using JSX syntax.

I'm not sure if there's a better place to put this documentation.

vishnugopal:
Am closing this now as it's quite old.

PR #239, opened 2021-03-03
Update mutation.mdx

sinap-fb:
currently searching for "ref" only results in reason-react related docs.  Maybe this would help also show a rescript ref result


ryyppy:
We need to fine tune algolia's indexing service at some point to pick up meta keywords instead. We'd prefer to not change the title.

ryyppy:
@sinap-fb thanks submitting a PR though! At some point we will add ref to our Syntax Lookup widget, so it will be actually better to use that tool for looking up language constructs

PR #242, opened 2021-03-07
Syntax lookup: let \""

kevanstannard:
Suggestion for #162

### DIFF
--- a/misc_docs/syntax/language_let_invalid.mdx
+++ b/misc_docs/syntax/language_let_invalid.mdx
@@ -1,0 +1,33 @@
+---
+id: "let-binding-invalid"
+keywords: ["let", "variable"]
+name: "let \\\"\""
+summary: "This is the `let binding` for invalid identifier names."
+category: "languageconstructs"
+---
+
+`let \""` is used to bind values to names that are invalid in ReScript, including capitalized names, names with special characters, and reserved keywords.
+
+### Example
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res
+let \"ComponentName" = "Header"
+let \"min-width" = 640
+let \"exception" = true
+```
+
+```js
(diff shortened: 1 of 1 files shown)


bobzhang:
This feature should not be exposed to end users. It may break lots of compiler invariants

kevanstannard:
Thanks @bobzhang 

@ryyppy what do you think? Should we remove the content here:

https://rescript-lang.org/docs/manual/latest/use-illegal-identifier-names

ryyppy:
Yeah the way this PR phrases it makes it look like a feature that users can safely use without any consequences. It is supposed to be an escape hatch only feature. 

Will discuss this internally. I believe this feature should be documented (but phrased differently), otherwise it might be abused. The goal is to document all language constructs available in the language.

I will adapt your texts later on, just keep the PR open.

kevanstannard:
Hi @ryyppy @bobzhang anything I can do to help resolve this one? I can have a go at adapting the text, or whatever else you have in mind. Happy to be guided by you.

misc_docs/syntax/language_let_invalid.mdx
```diff
@@ -0,0 +1,33 @@
+---
+id: "let-binding-invalid"
+keywords: ["let", "variable"]
+name: "let \\\"\""
+summary: "This is the `let binding` for invalid identifier names."
+category: "languageconstructs"
+---
+
+`let \""` is used to bind values to names that are invalid in ReScript, including capitalized names, names with special characters, and reserved keywords.
```
ryyppy, line 9:
```suggestion
`let \""` is used to bind values to names that are invalid in ReScript, including capitalized names, names with special characters, and reserved keywords.

**Warning: ** This escape hatch is almost exclusively used for interop with JavaScript. If you abuse this, many of the compiler guarantees will go away.
```

PR #244, opened 2021-03-08
Syntax lookup: if / else

kevanstannard:
Suggestion for #162 

* Moved existing `if / else` into **Language Constructs**
* Tweaked copy
* Removed **ControlFlow** category


ryyppy:
love it. thanks!

PR #247, opened 2021-03-08
Syntax lookup: function / uncurried function

kevanstannard:
Suggestion for #162 

* Add "function"
* Add "uncurried function"
* Remove old `function_uncurried.mdx` file.


ryyppy:
I am still wondering if my proposal of putting `() => {}` and `(.) => {}` as a keyword is the right choice, but we are still able to change that if ppl find that confusing

ryyppy:
Thanks for your work!

PR #259, opened 2021-03-15
Update .github/workflows/pull-request.yml

smorimoto:
- v2 is more fast
- Node.js 14 is the latest LTS
- You should use `--frozen-lockfile` in CI (Read the yarn documentation for details)


ryyppy:
Yeah, looks sensible. Thanks!

PR #267, opened 2021-03-20
Add ReScript-React migration guide

chenglou:


chenglou:
We need more info regarding migrating deps

PR #269, opened 2021-03-22
rescript-react style docs

ryyppy:


pages/docs/react/latest/styling.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+// Use the obj["key"] syntax to access any classname within our object
+let app = <div className={styles["root"]} />
+```
+
+**Note:** `{..}` is an open JS object type, which means the type checker will not type check correct classname usage. If you want to enforce compiler errors, replace `{..}` with a concrete JS object type, such as `{"root": string}`.
+
+
+## CSS Utility Libraries
+
```
rickyvetter, line 82: I definitely like the idea of showing that using common css libs is straightforward but I also think that just posting these two might show like a super strong endorsement when really they are some examples that we have used recently. Maybe a small disclaimer that there are other options out there as well?

```suggestion

There are a number of third party libraries for styling React components and you might be comfortable with a specific setup. If they work in JS then they almost certainly work in ReScript. Below, we've shared some ideas for working with popular libraries. If you're interested in working with one you don't see here, search the [package index](https://rescript-lang.org/packages) or post in [the forum](https://forum.rescript-lang.org)

```

pages/docs/react/latest/styling.mdx
```diff
@@ -0,0 +1,207 @@
...
+```res
+let container = Emotion.rawCss(`
+  color: #fff;
+  background-color: red;
+`)
+
+let app = <div className={container} />
+```
+
+For more CSS-in-JS ideas, you can also check out related discussions, or start a new topic on our [forum](https://forum.rescript-lang.org).
```
rickyvetter, line 207: Oh yeah you've got the link here too, just trying to make it more discoverable above.

ryyppy: I think I will remove that sentence, since we already have it in the beginning!

ryyppy: a wait, just realized this is for the CSS-in-JS part

PR #270, opened 2021-03-23
Add precisions about CSS-in-JS

bloodyowl:
- Shows examples where we hoist the output className (vs inlining it in props, which degrades performances for no reason when in `render`)
- Add "organization" example where styles are grouped in a `Styles` module (like React Native)
- Feature a binding that lets users input raw CSS (quite popular)

### DIFF
--- a/pages/docs/react/latest/styling.mdx
+++ b/pages/docs/react/latest/styling.mdx
@@ -157,7 +157,7 @@ let container = Emotion.css({
 let app = <div className={container} />
 ```
 
-You can organize your module stylesheet in a React Native fashion using first-class modules!
+You can also use submodules to organize your styles more easily:
 
 ```res
 module Styles = {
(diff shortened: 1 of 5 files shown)


pages/docs/react/latest/styling.mdx
```diff
@@ -148,12 +149,27 @@ The most minimalistic approach is to create simple bindings to e.g. [`emotion`](
...
-  className={Emotion.css({
+let container = Emotion.css({
+  "color": "#fff",
+  "backgroundColor": "red"
+})
+
+let app = <div className={container} />
+```
+
+You can organize your module stylesheet in a React Native fashion using first-class modules!
```
ryyppy, line 160: There are a few things that make this sentence too complex IMO:
- React Native -> We never mentioned it, nor do I think it should be mentioned to keep it lean
- first-class-modules -> I instantly thought about FCM, what you meant is a submodule, right?

```suggestion
You can also use submodules to organize your styles more easily:
```

ryyppy:
Awesome, thanks!

PR #271, opened 2021-03-25
Update latest doc to 9.0

namenu:
It's been almost 2 months since 9.0 rolled out; this is a good time to update the latest document to 9.0.

https://rescript-lang.org/blog/release-9-0

- [x] Cleaner Polyvariant Syntax
- [x] `when` -> `if` in pattern matching

### DIFF
--- a/compilers/package-lock.json
+++ b/compilers/package-lock.json
@@ -1,13 +1,36 @@
 {
   "name": "compilers",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "1.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "rescript-901": "npm:bs-platform@9.0.1"
+      }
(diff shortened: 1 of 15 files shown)


ryyppy:
Hey! Thanks for the initiative.

The reason why didn't upgrade this just yet is because we first need to:
- Upgrade the latest docs to the 8.4 version (add missing features that were introduced in the meantime, like monorepo-support etc.)
- Then copy that `latest` to a `v8` directory to signify that this is the latest up to date resources for all compiler versions in the v8 branch
- Then go forward with the `latest` being `v9` and address all the changes that are in there

Additionally, we wanted to start introducing `since` or `previously` notes to help users on different versions understand when a certain feature was made available.

It's the first time since the original Reason -> ReScript transition that we are doing such a "docs version fork", so the processes need to be carefully coordinated.

namenu:
Fair enough. I just thought that not having a doc for the latest major release is a bigger problem than preparing 8.4.
I'll give it another try when the core team finishes the transition.

PR #278, opened 2021-03-28
Remove some old docs that are no longer relevant

chenglou:


chenglou:
cc @ryyppy for this one. Just in case.

PR #281, opened 2021-03-29
Update safari favicon

chenglou:
IIRC safari does a heuristics on the favicon png to see if the edges are filled colors. If so, display at regular size. If not (like in our case, where we already have rounded borders in the icon), then display at smaller size

We want the full sized icon. So this icon fills the edges.


chenglou:
Screen Shot 2021-03-29 at 7 17 03 AM

After vs before

PR #289, opened 2021-04-03
Update 2021-03-03-rescript-association-rebranding.mdx

clarus:
If I am not mistaking, I think this is the Tezos Foundation rather than the Tezos Software Foundation.


ryyppy:
that's correct... thanks for correcting!

PR #293, opened 2021-04-07
Upgrade docs to 8.3, 8.4 and then to 9.x

ryyppy:
Unrelated to version:

- (done)Add `open!` docs

8.3:
- (done) remove `bs.` prefix from the attributes
- (done) Allow more character sets in module names (.something.res)
- (done) Fix Js.Array / Js.Array2 (https://github.com/rescript-lang/rescript-compiler/pull/4597/files)

8.4:
- (done) monorepo support feature [pinned deps](https://rescript-lang.org/blog/bucklescript-release-8-4#introducing-pinned-dependencies) 
- (done) introduce a new warning 109: toplevel expression is expected to have type unit It is turned on as warn-error by default.  This warning is introduced to avoid partial application errors in a curried language

Upgrade to 9.0:
- (done) Implement light weight syntax for poly-variants
- (done) New external stdlib improvement (https://rescript-lang.org/blog/release-9-0#new-external-stdlib-configuration)
- (done) when -> if (https://rescript-lang.org/blog/release-9-0#when---if)

### DIFF
--- a/_blogposts/2020-12-07-release-8-4.mdx
+++ b/_blogposts/2020-12-07-release-8-4.mdx
@@ -55,30 +55,9 @@ dependent's command line flags for building binary artifacts. Such strategy bene
 
 ### Introducing `pinned-dependencies`
 
-To make `bsb -make-world` more practical, we also introduce a new concept called: `pinned-dependencies`.
-In general, packages are classified as three categories:
+To make `bsb -make-world` more practical for e.g. monorepo setups (`lerna`, `yarn workspaces`, etc.), we introduced a new concept called `pinned-dependencies`. A pinned dependency allows us to automatically rebuild packages that are pinned by a toplevel package, like your final webapp.
 
-- toplevel
-  - warnings reported
-  - warn-error respected
-  - build dev dependency
-  - run custom rules
-  - package-specs like ES6/CommonJS overrides all its dependencies
-- pinned dependencies
-  - warnings reported
-  - warn-error respected
-  - build dev dependency
-  - run custom rules
-- normal dependencies
-  - warnings, warn-error ignored
-  - ignore dev directories
-  - ignore custom generator rules
-
-Previously, we only had `toplevel` and `normal dependencies`, by introducing `pinned-dependencies`,
-such package will be built mostly like `toplevel` packages.
-
-The usage is quite simple, add `pinned-dependencies` in the toplevel package's `bsconfig.json`.
-Note such field only make sense in toplevel's configuration.
+Please refer to our [Monorepo Setup](/docs/manual/latest/build-monorepos) docs for more information on how to use `pinned-dependencies` with different monorepo setups.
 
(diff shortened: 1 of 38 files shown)


ryyppy:
\cc @namenu

pages/docs/manual/latest/build-external-stdlib.mdx
```diff
@@ -0,0 +1,66 @@
...
+        console.log(num);
+        
+      }));
+```
+
+</CodeTab>
+
+The JavaScript output above was compiled with an `es6` target, but will also work with `commonjs`.
+
+**Important:** When using this option, you need to make sure that the version number of `bs-platform` and `@rescript/std` matches with the same version number in your `package.json` file, otherwise you'll eventually run into runtime problems due to mismatching stdlib behavior!
```
namenu, line 62: “this option” is not about `es6` or `commonjs`, right?
IMO the above sentence can be removed in this context for clearer meaning 🙂

_blogposts/2020-12-07-release-8-4.mdx
```diff
@@ -55,30 +55,9 @@ dependent's command line flags for building binary artifacts. Such strategy bene
...
-  - ignore custom generator rules
-
-Previously, we only had `toplevel` and `normal dependencies`, by introducing `pinned-dependencies`,
-such package will be built mostly like `toplevel` packages.
-
-The usage is quite simple, add `pinned-dependencies` in the toplevel package's `bsconfig.json`.
-Note such field only make sense in toplevel's configuration.
+To make `bsb -make-world` more practical for e.g. monorepo setups (`lerna`, `yarn workspaces`, etc.), we introduced a new concept called `pinned-dependencies`. A pinned dependency allows us to automatically rebuild packages that are pinned by a toplevel package, like your final webapp.
+
+Please refer to our [Monorepo Setup](/docs/manual/latest/build-monorepos) docs for more information on how to use `pinned-dependencies` with different monorepo setups.
```
bobzhang, line 60: My understanding of monorepo is google or facebook style's monorepo which is one repo one version control system. I am not sure why people call this kind of setup monorepo, we do have good support for such monorepo since our build system is scalable.

For multiple packages development (pinned dependencies), our support is best effort, the user experience is not as good as "monorepo support" (as mentioned above), I am a bit worried about that when our users read such section, they assume we have very good support for their understanding of "monorepo". This might cause some user expectation mismatch

ryyppy: Okay, so the term we are looking for is "Multiple Packages Development" instead of monorepo?
Yes, the term "monorepo" is a little convoluted in the JS world... I explicitly mentioned that the "original monorepo" use-case is well supported in the monorepo setup page

pages/docs/manual/latest/build-external-stdlib.mdx
ryyppy: Simplified this part

PR #295, opened 2021-04-09
Fix typo in workspaces example

nkrkv:
Looks like the original idea of the example was some `legacy` package which later became `myplugin`


ryyppy:
woops, leftover from my copy paste project. Thanks 🙏

PR #300, opened 2021-04-13
Fix toLocaleTimeString Typo using toLocaleString in examples

dan003400:
Fix typo in docs for `toLocaleTimeString` where docs show examples using `toLocaleString`.


ryyppy:
Woah. Thanks!

dan003400:
Any reason the Dom.Storage `setItem` was not added?

ryyppy:
Phew, probably the wrong place to have this discussion. My wild guess is ... ppl just didn't add it, because most of us write their own bindings because the builtin ones are pretty inconsistent.

dan003400:
I was only asking because it was also included in this PR. It does look like its already included, I just noticed its not documented currently.

https://github.com/dan003400/rescript-lang.org/commit/47ccb0eb64a0a2af3f231edff4fe5502d1deaa5c

ryyppy:
right, this was a while ago!

PR #315, opened 2021-05-03
Syntax lookup: Open

kevanstannard:
Suggestion for #162

**PR should be editable, so feel free edit**

### DIFF
--- a/misc_docs/syntax/language_open.mdx
+++ b/misc_docs/syntax/language_open.mdx
@@ -1,0 +1,35 @@
+---
+id: "open"
+keywords: ["open", "module"]
+name: "open"
+summary: "This is the `open` keyword."
+category: "languageconstructs"
+---
+
+`open` is used to "open" a module so  we can refer to the module's contents without needing to prepending the module's name.
+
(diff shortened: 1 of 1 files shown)


misc_docs/syntax/language_open.mdx
```diff
@@ -0,0 +1,35 @@
+---
+id: "open"
+keywords: ["open", "module"]
+name: "open"
+summary: "This is the `open` keyword."
+category: "languageconstructs"
+---
+
+`open` is used to "open" a module so  we can refer to the module's contents without needing to prepending the module's name.
```
ryyppy, line 9:
```suggestion
`open` is used to expose all values, types, modules, etc of a module in the current scope. This is useful whenever you want to use a module's functionality without typing out the module name over and over again.
```

misc_docs/syntax/language_open.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+In some cases, `open` will cause a warning due to existing identifiers and types being redefined by the opened module. In these cases, we can use `open!` instead which will suppress the warnings.
```
ryyppy, line 11:
```suggestion
In some cases, `open` will cause a "shadow warning" due to existing identifiers and types being redefined by an `open`ed module. You can explicitly ignore that warning by using an `open!` statement instead.
```

PR #323, opened 2021-05-05
Syntax lookup: Type parameter

kevanstannard:
Suggestion for #162

Simple explanation for a type parameter.

The Syntax Widget point for this one suggests including `+'a` for covariance, but I'm not very familiar with variance yet, so I haven't included it. 

PR should be editable, please feel free edit.


ryyppy:
This looks good.

PR #328, opened 2021-05-06
Remove custom blog category

chenglou:
Some of the categorization of our posts aren't clear, and some would
require multiple categories. Given our pace, I don't think there's any
need to filter more finely based on sometime inaccurate categories. The
default "all" and "archived" are enough


ryyppy (approved):

ryyppy:
lgtm

PR #334, opened 2021-05-06
Update embed-raw-javascript.mdx

idkjs:
update add example so it works.

### DIFF
--- a/pages/docs/manual/latest/embed-raw-javascript.mdx
+++ b/pages/docs/manual/latest/embed-raw-javascript.mdx
@@ -40,7 +40,7 @@ While `%%raw` lets you embed top-level raw JS code, `%raw` lets you embed expres
 <CodeTab labels={["ReScript", "JS Output"]}>
 
 ```res example
-let add = %raw(`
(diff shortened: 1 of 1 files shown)


ryyppy:
I am confused. The original example [works just fine](https://rescript-langorg-7xv0hjoyn-rescript-association.vercel.app/try?code=DYUwLgBAhgJjEF4IFIBOUDuAKABgKAggDMBXAOwGMwBLAezKygBoIAjASggG8DCIL6AZ1qgAdMFoBzLACIAFiGATiqWgFsI6DBABSUAG5QAyhVTUADmACEM9gG5ehVOBKoy0CAGo2vAL54cdjw8HUFxKUY4LABGFgAmdnYgA), so what is there to fix?

Adding / removing a type annotation has 0 impact on the JS output.

Are you running on an old compiler version maybe?

chenglou:
We welcome PRs but it'd be nice if you can read what we've written on the forum about this: please provide your compiler version and a repro. This is a simple example that just doesn't repro at all.

PR #337, opened 2021-05-06
Clean up gitignore and yarn

chenglou:


.gitignore
```diff
@@ -11,17 +9,9 @@ node_modules/
 .next/
 index_data/*.json
 
-pages/docs/manual/v8.0.0/
```
chenglou, line 14: @ryyppy this isn't ignored

PR #347, opened 2021-05-08
Reflect submodule structure in Belt docs layout

Minnozz:
For clarity, `HashMapString` should be called `HashMap.String` since that is how you reference that module in ReScript. Same for other submodules.

### DIFF
--- a/pages/docs/manual/latest/api/belt/hash-map-int.mdx
+++ b/pages/docs/manual/latest/api/belt/hash-map-int.mdx
@@ -1,4 +1,4 @@
-# HashMapInt
+# HashMap.Int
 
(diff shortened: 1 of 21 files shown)


ryyppy:
hmmmm not too sure about that change. I want to keep this untouched until we automatically generate the new docs from the actual source code. Back then we translated those headers 1 by 1 from odoc generated html

Minnozz:
Is generating the docs from source tracked in an issue somewhere? I couldn't find it.

ryyppy:
Just created one here: #359

Minnozz:
Cool! I'll close this PR then.

PR #348, opened 2021-05-08
Syntax lookup: Send pipe decorator

kevanstannard:
Suggestion for #162

* Also a few tweaks to the Send decorator page
* It looks like Send pipe still needs the `bs.` prefix.

**PR should be editable, please feel free edit.**


ryyppy:
Thanks!

PR #350, opened 2021-05-08
Syntax lookup: Warning decorator

kevanstannard:
Suggestion for #162 

**PR should be editable, please feel free edit.**

### DIFF
--- a/misc_docs/syntax/decorator_expression_warning.mdx
+++ b/misc_docs/syntax/decorator_expression_warning.mdx
@@ -2,7 +2,7 @@
 id: "expression-warning-decorator"
 keywords: ["warning", "decorator"]
 name: "@warning"
-summary: "This is the `expression warning` decorator."
+summary: "This is the `@warning` decorator."
 category: "decorators"
 ---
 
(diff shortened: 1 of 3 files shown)


kevanstannard:
@ryyppy yes, the different syntaxes seem to be working. I tried the following:

```res
@warning("-32")
let x = 0 // No warning here
let x = 1

let y = 0 // "unused value y" warning here
let y = 1
```

As with your comment on the  `@deprecated` PR, what do you think about making these separate pages for Syntax Lookup?

Anything else to consider?

ryyppy:
Two separate pages are fine. Just let's make it clear that the decorator applies to an expression / module.

kevanstannard:
@ryyppy PR updated, see what you think.

Also, could I double check something? In the PR I wrote that the decorators are use to _enable or disable_ warnings. Is it correct that they can _enable_ warnings, or are they exclusively used for disabling warnings?

ryyppy:
> Is it correct that they can enable warnings, or are they exclusively used for disabling warnings?
You can do both: Enable (`@warning("+27")`) or Disable (`@warning("-27")`)

kevanstannard:
Thanks @ryyppy 

I've made some tweaks, see what you think.

bobzhang:
My concern with issue is that this is a buggy feature which has its use case but very hard to implement it correctly.

kevanstannard:
Syntax Lookup @@warning is documented now, so let's close this one off.

PR #361, opened 2021-05-13
Add %external() class feature detection example

### LINKED ISSUE #351: Document %external(\"SomeValue")
We have `%external` documented on the [Special Global Values](https://rescript-lang.org/docs/manual/latest/bind-to-global-js-values#special-global-values) but I'm not sure if we have the `%external(\"SomeValue")` syntax documented?

Just noticed this on a [recent forum post](https://forum.rescript-lang.org/t/feature-detection-in-the-browser-with-capital-leaders-illegal-identifiers/1582).

I might have missed it in the docs, otherwise is this something we would like to document?

I am happy to contribute a PR if you think this is useful to document.

Thanks

kevanstannard:
Fixes #351

As discussed in the issue.

**Should be editable, feel free to edit**

### DIFF
--- a/pages/docs/manual/latest/bind-to-global-js-values.mdx
+++ b/pages/docs/manual/latest/bind-to-global-js-values.mdx
@@ -142,4 +142,36 @@ if (match !== undefined) {
 
 </CodeTab>
 
+This approach may also be used for JS class feature detection.
+
+Note that here we are using the `\""` syntax which allows us to [use a capitalized identifier name](./use-illegal-identifier-names).
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type audioContext
+
+@new external createAudioContext: unit => audioContext = "AudioContext"
+
+let createAudioContext: unit => option<audioContext> = () => {
+  switch %external(\"AudioContext") {
+  | Some(_) => createAudioContext()->Some
+  | None => None
+  }
+}
+```
+
+```js
+var Caml_option = require("rescript/lib/js/caml_option.js");
+
+function createAudioContext(param) {
+  var match = typeof AudioContext === "undefined" ? undefined : AudioContext;
+  if (match !== undefined) {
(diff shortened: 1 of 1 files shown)


ryyppy:
\cc @chenglou

kevanstannard:
Hi @chenglou @ryyppy just following up.

I'm also mindful this is using the `\""` syntax ( referring to #242 and https://github.com/rescript-lang/rescript-compiler/issues/5269 )

Any preference on next steps for this issue?

bobzhang:
For this particular feature request, I am fine to generalize it so user can write `%external("AudioContext")`.
Is it generally useful & what's the benefit compared with using raw directly?

kevanstannard:
Thanks @bobzhang 

If we use raw, do you mean something like this?

```
type audioContext
@new external createAudioContext: unit => audioContext = "AudioContext"

let canUseAudioContext: bool = %raw(`typeof AudioContext === "function"`)
let audioContext = canUseAudioContext ? Some(createAudioContext()) : None
```

I personally would be fine to use that, but I'm happy to be guided by the team.

Just a little feedback on `%external()`. I was a bit confused how it worked at first, since `%external(a)` worked, but `%external(A)` didn't work, and `%external("A")` didn't work and ``%external(\"A")` did work. From this I realised that the naming is based on using valid identifiers.

Was your suggestion to generalize mean that we would always using a string? E.g. `%external("__DEV__")`, `%external("a")`,  `%external("A")`? That sounds like a helpful improvement, and also avoids using `\""`.

If that's the case, would it be better to not document this for now?

bobzhang:
> Was your suggestion to generalize mean that we would always using a string?
Yes

In my opinion, `external` here only has marginal benefit over using `raw`, so maybe we should stick to raw for simplicity

kevanstannard:
Thanks @bobzhang, in that case let's close this PR.

PR #368, opened 2021-05-15
[Syntax lookup] Add doc for `char` literal syntax

kellydanma:
Proposal for https://github.com/rescript-association/rescript-lang.org/issues/162
- `' '` (single quote = char)


misc_docs/syntax/language_char_literal.mdx
```diff
@@ -0,0 +1,34 @@
+---
+id: "char-literal"
+keywords: ["char"]
+name: "''"
```
ryyppy, line 4: Maybe add a whitespace, so it's easier to distinguish from a `"`?
```suggestion
name: "' '"
```

ryyppy:
Looking good, thanks!

PR #373, opened 2021-05-23
Type-o in the example script as rescript is a bin

anthonymuau:
The script is referencing node modules folder for rescript and not the bin file of rescript causing the script to fail.


ryyppy:
I generally think this command is unnecessary bc you can run ˋnpx rescript convert (some/dir]ˋ as well

PR #375, opened 2021-05-23
Highlighted UTF-8 / unicode compatibility gotchas

lietu:
I had skimmed through the docs, but when I ran into the issue with `""` strings blowing up with UTF-8 contents (`Ü`, `♥`) I had no clue what I could do about this. Gladly some helpful people on the ReasonML Discord helped out with it, but before that I had spent quite a while searching for answers on both the docs and search engines.

The docs briefly mention the issue casually in middle of a title called "String interpolation" - not great. The only mention of "utf-8" on that page takes you to a section which says `Note: Char doesn't support Unicode or UTF-8 and is therefore not recommended.` - making it seem unicode is not supported at all.

Several of the top results on search engines seem to also indicate that OCaml has no support for unicode and so ReasonML/ReScript also cannot support it.

This clear highlight at the top of the page should at least help with identifying a common critical issue which a lot of users are guaranteed to run into.

### DIFF
--- a/pages/docs/manual/latest/primitive-types.mdx
+++ b/pages/docs/manual/latest/primitive-types.mdx
@@ -10,6 +10,21 @@ ReScript comes with the familiar primitive types like `string`, `int`, `float`, 
 
 <!-- TODO: doc unit -->
 
+## Unicode compatibility
+
+⚠ **NOTE:** Strings in ReScript only support unicode in UTF-8 formatted files properly when wrapped in backticks:
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+let wrong = "I ♥ Günther"
+let right = `I ♥ Günther`
+```
+```js
+var wrong = "I \xe2\x99\xa5 G\xc3\xbcnther";
+var right = "I ♥ Günther";
+```
+
 ## String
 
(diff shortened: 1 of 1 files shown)


ryyppy:
Hey! Thanks for the PR. I will come back to this in a bit.

pages/docs/manual/latest/primitive-types.mdx
```diff
@@ -10,6 +10,21 @@ ReScript comes with the familiar primitive types like `string`, `int`, `float`,
...
+
+```res example
+let wrong = "I ♥ Günther"
+let right = `I ♥ Günther`
+```
+```js
+var wrong = "I \xe2\x99\xa5 G\xc3\xbcnther";
+var right = "I ♥ Günther";
+```
+
```
ryyppy, line 27:
```suggestion

<CodeTab/>
```

pages/docs/manual/latest/primitive-types.mdx
```diff
@@ -10,6 +10,22 @@ ReScript comes with the familiar primitive types like `string`, `int`, `float`,
...
+<CodeTab/>
```
ryyppy, line 28:
```suggestion
<CodeTab/>

```

fhammerschmidt:
Unicode is supported in "" as well for some time now. Closing.

PR #378, opened 2021-05-27
Invalid prop names needs to be snake_case

srikanthkyatham:
camelCasing for invalid prop names is not working, snake_case for invalid prop names seems to work on Rescript 9.1

meaning
ariaLabel is getting translated to ariaLabel
aria_label is getting translated to aria-label - which is desired


srikanthkyatham:
Wrong understanding

PR #381, opened 2021-05-31
Invalid props for Non DOM Components

srikanthkyatham:
Updating documentation for passing `aria-*` props to Non DOM components

### DIFF
--- a/pages/docs/react/latest/components-and-props.mdx
+++ b/pages/docs/react/latest/components-and-props.mdx
@@ -152,7 +152,7 @@ Check out the corresponding [Arrays and Keys](./arrays-and-keys) and [Forwarding
 
 Prop names like `type` (as in `<input type="text" />`) aren't syntactically valid; `type` is a reserved keyword in ReScript. Use `<input type_="text" />` instead. 
 
-For `aria-*` use camelCasing, e.g., `ariaLabel`. For DOM components, we'll translate it to `aria-label` under the hood.
+For `aria-*` use camelCasing, e.g., `ariaLabel`. For DOM components, we'll translate it to `aria-label` under the hood. For Non DOM components passing `aria-*` needs to be done like `data-*`.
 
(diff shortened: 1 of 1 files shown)


ryyppy:
Hm, I read this sentence it was more confusing to me than before.

We are separating between DOM elements `<div/>` and actual React components `<MyButton/>`. You want to make clear that for React component props, the `aria-` transformation doesn't apply, right?

I still don't understand the motivation for this, because if you are creating a component that does introduce a `~ariaLabel` prop, it would still need to apply that argument to a DOM element somewhere.

```res
@react.component
  let make = (~ariaLabel: string) => {
    <button ariaLabel> {"test"->React.string} </button>
  }
```

I don't see any scenario where this might be an issue?

srikanthkyatham:
Hi @ryyppy 


> Hm, I read this sentence it was more confusing to me than before.
> 
> We are separating between DOM elements `<div/>` and actual React components `<MyButton/>`. You want to make clear that for React component props, the `aria-` transformation doesn't apply, right?
> 

Yes your understanding is correct 

> I still don't understand the motivation for this, because if you are creating a component that does introduce a `~ariaLabel` prop, it would still need to apply that argument to a DOM element somewhere.
> 
> ```rescript
> @react.component
>   let make = (~ariaLabel: string) => {
>     <button ariaLabel> {"test"->React.string} </button>
>   }
> ```
> 
> The `<button>` element is doing the transform correctly, so can't see any scenario where this might be an issue?

The use case was pass aria-* props to React components from a ui library. In our case a company's ui library.

ryyppy:
> The use case was to pass aria-* props to React components from a ui library written in javascript. In our case a company's ui library.

This sounds like you are defining a `type Props = { "aria-label": string }` in your existing React component? This doesn't sound right to me for some reason.

ryyppy:
I will close this for now, feel free to initiate the discussion again if you still feel strongly about it.

PR #387, opened 2021-06-12
Fix incorrectly migrated package-lock.json

Minnozz:
`node_modules/.bin` was missing after a clean `npm install`

See npm/cli#2147

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ This is the official documentation platform for the [ReScript](https://rescript-
 ## System Requirements
 
 - `node@12.22.1` or higher (for ES6 module compat)
-- `npm@6` (package-lock v1)
+- `npm@7` (package-lock v2)
 
 ## Setup
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,6 +73,9 @@
         "resolve": "^1.3.2",
         "semver": "^5.4.1",
         "source-map": "^0.5.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/generator": {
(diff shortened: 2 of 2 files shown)


ryyppy:
we need to revisit this later, since we currently have a really weird, but working setup on our team's machines that we don't want to touch rn

Minnozz:
Could you explain the weird setup? 

I think the current `package-lock.json` is just buggy after being recently upgraded from lockfile v1 to v2 (https://github.com/rescript-association/rescript-lang.org/commit/b1ecde64fbe2152370fdae26cff6cb3e7fd749b3#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519), causing the `rescript` and other binaries to be unavailable in `npm run` scripts unless you had already run an `npm install` in your checkout before that commit.

ryyppy:
Right now you are supposed to use `node@12` and `npm@6` (this was not a typo).

npm@6 uses lockfile v1, which works perfectly fine on all our machines.
We experienced problems with lockfile version 2 though, which ended up so badly that Cheng was literally zipping up `node_modules` to be able to reproduce his rescript-lang.org setup.

Same for Maxim and Bettina's machine (and partly also mine).

Will have another look at it soon, right now it's too cumbersome to me to communicate yet another tooling version change for everyone.

Minnozz:
Right, I understand. Feel free to ignore this / leave this until later, but just to clear up any confusion:

I did not migrate the lockfile to v2 in this PR, that was done a month ago in this commit: https://github.com/rescript-association/rescript-lang.org/commit/b1ecde64fbe2152370fdae26cff6cb3e7fd749b3. I just updated the README to reflect the lockfile version that is currently in use. If you want to stay on `npm@6` it would be better to revert the lockfile in the repo to v1; I could do that if you want?

The team's problems with the v2 lockfile may be the same I encountered (missing binaries)? Both downgrading to v1 or re-upgrading to v2 (this PR) would fix that.

ryyppy:
Okay now I understand the problem. Will merge this and rebase the currently active branches.

PR #389, opened 2021-06-13
Change `React.useEffect1(fn, [])` to `React.useEffect0(fn)`

Minnozz:
No change in behaviour, but this aligns better with [the docs](https://rescript-lang.org/docs/react/latest/hooks-effect#effect-dependencies).


ryyppy:
could you also check in the js build artifacts?

Minnozz:
The build artifacts were identical since both calls compile to `React.useEffect(fn, [])` in JavaScript.

ryyppy:
perfect. Then it's safe to merge I guess. Thanks!

PR #390, opened 2021-06-13
Improve documentation for `@as("str")` and `@as("0")` decorator in manual and syntax widget

### LINKED ISSUE #189: missing docs for record renaming
```res
type t = {
  @as("X") x: int,
  @as("Y") y: string,
}
```
and a special case for encoding array
```res
type t = {
  @as("0") x : int,
  @as("1") y : int
}
```

Minnozz:
Fixes #189


ryyppy:
this looks nice. thanks!

PR #391, opened 2021-06-14
Landingpage responsive details

bettistein:


kanishka-work:
Here are two issues that I observed on Firefox mobile:

Screenshot_20210625-084758
Screenshot_20210625-084717

PR #395, opened 2021-06-19
Document warning numbers

### LINKED ISSUE #190: Add available warning numbers for @@warning()
We already map our warnings numbers in our playground (check the "Settings" tab, and try to add a warning to see the number recommendations).

The same data could be used to list all the available warnings in our Attribute / Decorator docs. Might be a little tricky to list them all, so might be worth making the list interactive.

Minnozz:
Fixes #190

### DIFF
--- a/data/sidebar_manual_latest.json
+++ b/data/sidebar_manual_latest.json
@@ -56,7 +56,8 @@
     "build-external-stdlib",
     "build-pinned-dependencies",
     "interop-with-js-build-systems",
-    "build-performance"
+    "build-performance",
+    "compiler-warnings"
   ],
   "Guides": [
     "converting-from-js"
--- a/data/sidebar_manual_latest.json
+++ b/data/sidebar_manual_latest.json
@@ -57,7 +57,7 @@
     "build-pinned-dependencies",
     "interop-with-js-build-systems",
     "build-performance",
-    "compiler-warnings"
+    "warning-numbers"
   ],
   "Guides": [
     "converting-from-js"
--- a/misc_docs/syntax/decorator_module_warning.mdx
+++ b/misc_docs/syntax/decorator_module_warning.mdx
@@ -1,0 +1,27 @@
+---
+id: "module-warning-decorator"
+keywords: ["warning", "decorator"]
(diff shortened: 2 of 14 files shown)


ryyppy:
This looks generally pretty good.

A few things:
- Not sure if we should call the sidebar navigation item `Compiler Warnings`... sounds more like an FAQ collection of error messages (problem -> solution). Maybe `Warning Numbers` would be better?
- I noticed that our Warnings collection has deviated from upstream (still a bunch of BuckleScript words in there). Here's the most up to date list: [warnings.ml](https://github.com/rescript-lang/ocaml/blob/2930b685e73fd60a63181a939ba641f63b50c588/utils/warnings.ml#L667-L753)

Probably unrelated to this list, but I also wanted to have a simple text field to fuzzy find particular warning numbers by description. In a similar manner how we did it with the Syntax Lookup Widget (using the Fuse.res bindings), but this can be done in another task.

ryyppy:
Nice! One last thing: 

Can we change the link within [this section](https://rescript-langorg-mk5t2ikrt-rescript-association.vercel.app/docs/manual/latest/build-configuration#warnings) to point to the new warning-number list?

That will be the last change, otherwise it looks perfect!

Minnozz:
We might want to link to the new warning list (or embed `<WarningTable />`) on some more pages: https://github.com/rescript-association/rescript-lang.org/issues/190#issuecomment-860908936

ryyppy:
> We might want to link to the new warning list 

Yeah, in the attribute page we could add a link to the warning list in the description here:
image

`You can find a full list of all available warnings [here](...)`

> most useful would be the syntax widget I think

I'd not include the whole table into different pages... should be enough to link to it, no? In the Syntax widget we use the `See` section to link to canonical resources

Minnozz:
I've added some more links to and from the Warning Numbers page, and I also added the `@@warning` decorator to the syntax widget. I'm not entirely sure if I'm correct that it operates on the module level; somewhere else it is stated that it operates on the file.

ryyppy:
this is looking sleek. Thanks!

PR #404, opened 2021-06-25
[Syntax lookup] Add doc for explicit `string` interpolation

kellydanma:


ryyppy:
Probably a mistake from my side for defining it as a `explicit string interpolation` syntax... it should actually just be called `string interpolation` syntax.

The older formats like j\`\` should not be used anymore anyways.

misc_docs/syntax/language_explicit_string_interpolation.mdx
```diff
@@ -0,0 +1,34 @@
+---
```
ryyppy, line 1: File should be called `language_string_interpolation.mdx`

misc_docs/syntax/language_explicit_string_interpolation.mdx
```diff
@@ -0,0 +1,34 @@
...
+id: "explicit-string-interpolation"
```
ryyppy, line 2:
```suggestion
id: "string-interpolation"
```

misc_docs/syntax/language_explicit_string_interpolation.mdx
```diff
@@ -0,0 +1,34 @@
...
+keywords: ["string", "interpolation"]
+name: "``"
+summary: "This is the explicit `string` interpolation syntax."
```
ryyppy, line 5:
```suggestion
summary: "This is the `string interpolation` syntax."
```

misc_docs/syntax/language_explicit_string_interpolation.mdx
```diff
@@ -0,0 +1,34 @@
...
+category: "languageconstructs"
+---
+
+An explicit `string` interpolation is composed of two backticks. It works like JavaScript's backtick string interpolation without the need to escape special characters.
```
ryyppy, line 9:
```suggestion
A `string interpolation` is composed of two backticks. It works like JavaScript's backtick string interpolation without the need to escape special characters.
```

ryyppy (changes requested):

PR #406, opened 2021-06-26
fix #399

### LINKED ISSUE #399: some warnings flags are not precise in rescript

bobzhang:
This is a small commit to remove the warning numbers which does not make sense in rescript.

Ideally we should make it better in two ways:

- More descriptions than just one line meaning
- Make it linkable (provide an anchor) so that the compiler can emit the link point here


ryyppy:
Are these warnings also removed in the compiler itself?

> Make it linkable (provide an anchor) so that the compiler can emit the link point here

Why would I link to the same one line description? Are you talking about an error faq with problem description and possible solutions to fix the error?

anyways, lgtm

bobzhang:
> Are these warnings also removed in the compiler itself?

Yes

> Why would I link to the same one line description?

If we can provide an anchor like this `http://rescript-lang.org/errors#33`, the compiler can emit that so people can click it for more details

PR #407, opened 2021-06-28
change vistorCount to visitorCountMsg

lhao03:
I was trying to build this example but got a type error.

image

I believe vistorCount should be visitorCountMsg. 

I have also run `npm test` and no tests are failing.

image

The corresponding JS Output is correct to my knowledge.


ryyppy:
thanks!

PR #410, opened 2021-07-08
docs: fix syntax problems

l2dy:
* `'children` is a type parameter, not `'children'`.
* `{. name: string, ...}` is Reason syntax for object type declaration. I think `{. "name": string, ...}` was meant to be https://rescript-lang.org/docs/manual/latest/object#type-declaration.


ryyppy:
thanks!

PR #424, opened 2021-08-15
Fix grammatical errors in pipe.mdx, and make some subjective wording tweaks

jamespwilliams:
I've made two commits here. The first fixes things that I think are objectively grammatically wrong. The second makes some more subjective changes to the text. If you only want the first, let me know and I'll remove the second commit.


bobzhang:
lgtm, thanks

PR #426, opened 2021-08-20
Fixing the reduceWithIndex documentation

srikanthkyatham:
The order of the parameters documentation was wrong, reduceWithIndex callback will receive the accumulator first, item from the array and then the index of the item

### DIFF
--- a/pages/docs/manual/latest/api/belt/array.mdx
+++ b/pages/docs/manual/latest/api/belt/array.mdx
@@ -765,7 +765,7 @@ let reduceWithIndex: (array<'a>, 'b, ('b, 'a, int) => 'b) => 'b
 
 `reduceWithIndex(xs, f)`
 
-Applies `f` to each element of `xs` from beginning to end. Function `f` has three parameters: the item from the array and an “accumulator”, which starts with a value of `init` and the index of each element. `reduceWithIndex` returns the final value of the accumulator.
+Applies `f` to each element of `xs` from beginning to end. Function `f` has three parameters: an “accumulator”, which starts with a value of `init` and the item from the array and the index of each element. `reduceWithIndex` returns the final value of the accumulator.
 
 ```res example
 Belt.Array.reduceWithIndex([1, 2, 3, 4], 0, (acc, x, i) => acc + x + i) == 16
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/api/belt/array.mdx
```diff
@@ -765,7 +765,7 @@ let reduceWithIndex: (array<'a>, 'b, ('b, 'a, int) => 'b) => 'b
 
 `reduceWithIndex(xs, f)`
```
ryyppy, line 766: Isn't this definition wrong as well?

```suggestion
`reduceWithIndex(arr, init, f)`
```
```suggestion
`reduceWithIndex(xs, f)`
```

Unrelated: always gets me that the array parameter is called `xs`... what does that even mean.

pages/docs/manual/latest/api/belt/array.mdx
```diff
@@ -765,7 +765,7 @@ let reduceWithIndex: (array<'a>, 'b, ('b, 'a, int) => 'b) => 'b
...
 
-Applies `f` to each element of `xs` from beginning to end. Function `f` has three parameters: the item from the array and an “accumulator”, which starts with a value of `init` and the index of each element. `reduceWithIndex` returns the final value of the accumulator.
+Applies `f` to each element of `xs` from beginning to end. Function `f` has three parameters: an “accumulator”, which starts with a value of `init` and the item from the array and the index of each element. `reduceWithIndex` returns the final value of the accumulator.
```
ryyppy, line 768:
```suggestion
Applies `f` to each element of `arr` from beginning to end. Function `f` has three parameters: an "accumulator", which starts with a value of `init` and the item from the array and the index of each element. `reduceWithIndex` returns the final value of the accumulator.
```

pages/docs/manual/latest/api/belt/array.mdx
```diff
@@ -765,7 +765,7 @@ let reduceWithIndex: (array<'a>, 'b, ('b, 'a, int) => 'b) => 'b
...
 
 ```res example
 Belt.Array.reduceWithIndex([1, 2, 3, 4], 0, (acc, x, i) => acc + x + i) == 16
```
ryyppy, line 771:
```suggestion
Belt.Array.reduceWithIndex([1, 2, 3, 4], 0, (acc, value, i) => acc + value + i) == 16
```

pages/docs/manual/latest/api/belt/array.mdx
srikanthkyatham: Yes that is correct reduceWithIndex is also wrong. your suggestion looks way better

pages/docs/manual/latest/api/belt/array.mdx
srikanthkyatham: xs -> arr looks much better while reading the docs

srikanthkyatham:
I will report any other findings, I am practically living on API docs nows a days.

ryyppy:
thanks!

PR #430, opened 2021-08-26
fix codeexample label overlap

vasco3:
The problem is that the background from the language label is hiding part of the code example

image

I propose making transparent the background of the label so it's not that bad

image


ryyppy:
Yeah i knew this is gonna happen... well, that is not an ideal solution, but it's a good fix for now!
thanks

PR #433, opened 2021-09-01
fix(ui): mobile navigation overlay

aspeddro:
Open Menu `...` and touch in `Docs`:

rescript-lang org_

Now, `Docs` overlays the Menu `...`


ryyppy:
Ideally, when a subnav is opened and the ... pressed, the other overlay should automatically close.. but this change is good! thanks.

PR #443, opened 2021-09-08
[Fix] Correct the typo "re" to "res"

saseungmin:
- Fix Typo comment: re => res in ReScript React section.


ryyppy:
BAM! Get all the `.re`s out of here 🔥

PR #444, opened 2021-09-09
Syntax Lookup: Extension Points

kevanstannard:
Suggestion for #162 

First pass at adding Extension Point docs to Syntax Lookup.

Feel free to edit.


misc_docs/syntax/extension_embed_raw_expression.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+```js
+var canUseCanvas = function canUseCanvas() {
+  return !!document.createElement("canvas").getContext;
+};
+```
+
+</CodeTab>
+
+See `%%raw` for embedding top level blocks of JavaScript code rather than expressions.
```
ryyppy, line 29: I guess we can directly refer to that as well!

```suggestion
See [`%%raw`](/syntax-lookup#embed-raw-top-level) for embedding top level blocks of JavaScript code rather than expressions.
```

misc_docs/syntax/extension_embed_raw_expression.mdx
```diff
@@ -0,0 +1,34 @@
+---
+id: "embed-raw-expression"
```
ryyppy, line 2: Why is it called `embed-*`? Wouldn't be `raw-expression` be enough?

I guess it's because you are embedding raw JS code... i am not sure if the extra term helps in the understanding though.

misc_docs/syntax/extension_embed_raw_top_level.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+function greet(m) {
+  console.log(m);
+}
+
+greet(message);
+```
+
+</CodeTab>
+
```
ryyppy, line 36:
~~~suggestion

It's also very useful to do imports with side-effects like this:

```res
%%raw(`import "main.css"`)
```

~~~

misc_docs/syntax/extension_embed_raw_top_level.mdx
```diff
@@ -0,0 +1,43 @@
...
+See `%raw` for embedding JavaScript expressions rather than top level blocks of code.
```
ryyppy, line 37:
```suggestion
See [`%raw`](/syntax-lookup#embed-raw-expression) for embedding JavaScript expressions rather than top level blocks of code.
```

misc_docs/syntax/extension_identity.mdx
```diff
@@ -0,0 +1,31 @@
+---
+id: "identity"
+keywords: ["identity", "external", "type", "convert"]
+name: "%identity"
+summary: "This is the `identity` extension point."
+category: "extensionpoints"
+---
+
+`%identity` is used with `external` to convert a value from one type to another type.
```
ryyppy, line 9:
```suggestion
`%identity` is used with `external` to do an **unsafe conversion** of a value from one type to another type.
```

misc_docs/syntax/extension_regular_expression.mdx
```diff
@@ -0,0 +1,29 @@
+---
+id: "regular-expression"
+keywords: ["regular", "expression", "re"]
+name: "%re"
+summary: "This is the `regular expression` extension point."
+category: "extensionpoints"
+---
+
+`%re` is used to insert JavaScript regular expressions.
```
ryyppy, line 9:
```suggestion
`%re` is used to create JavaScript regular expressions.
```

misc_docs/syntax/extension_embed_raw_top_level.mdx
```diff
@@ -0,0 +1,43 @@
+---
+id: "embed-raw-top-level"
```
ryyppy, line 2: I'd kinda be in favour calling it `raw-top-level-expression`

```suggestion
id: "raw-top-level-expression"
```

ryyppy (commented):
Looking really sharp 👍  

Made some comments, might need some file renaming, and some commit suggestion batching. All in all I like it.

misc_docs/syntax/extension_embed_raw_expression.mdx
kevanstannard: Thanks for the feedback, I was undecided with this one, but I think your point about it not helping with understanding makes sense.

I've changed to the following:

`%raw` == "JavaScript expression" extension point
`%%raw` == "Top level JavaScript" extension point

Also, since the syntax uses the term `raw`. Do you think should be included in the name?

`%raw` == "Raw JavaScript expression" extension point
`%%raw` == "Raw top level JavaScript" extension point

I was thinking that "JavaScript" and "raw" here are almost synonymous in this context.

Happy to refine further.

misc_docs/syntax/extension_embed_raw_expression.mdx
kevanstannard: Links like this within the Syntax Lookup are not working for me at the moment. Maybe the router is not detecting the URL hash changes, or something like that? What do you think, would it be worth including this now in anticipation of updating the Syntax Lookup routing behaviour?

misc_docs/syntax/extension_embed_raw_top_level.mdx
kevanstannard: Great, I've included this change 👍

misc_docs/syntax/extension_raw_expression.mdx
```diff
@@ -0,0 +1,34 @@
+---
+id: "raw-expression"
+keywords: ["javascript", "raw", "expression"]
+name: "%raw"
+summary: "This is the `JavaScript expression` extension point."
```
ryyppy, line 5: Trying to be consistent here

```suggestion
summary: "This is the `raw expression` extension point."
```

misc_docs/syntax/extension_raw_expression.mdx
```diff
@@ -0,0 +1,34 @@
...
+category: "extensionpoints"
+---
+
+`%raw` is used to embed JavaScript expressions.
```
ryyppy, line 9: Maybe this makes it clearer why it's called `raw`.. we want to embed `raw JS` basically.

```suggestion
`%raw` is used to embed "raw JavaScript expressions".
```

misc_docs/syntax/extension_raw_top_level.mdx
```diff
@@ -0,0 +1,57 @@
+---
+id: "raw-top-level"
```
ryyppy, line 2: Oops, last try

```suggestion
id: "raw-top-level-expression"
```

misc_docs/syntax/extension_raw_top_level.mdx
```diff
@@ -0,0 +1,57 @@
...
+keywords: ["javascript", "raw"]
+name: "%%raw"
+summary: "This is the `top level JavaScript` extension point."
```
ryyppy, line 5:
```suggestion
summary: "This is the `raw top level expression` extension point."
```

misc_docs/syntax/extension_embed_raw_top_level.mdx
kevanstannard: Lol, I was unsure about naming for this one.

I left a [comment about naming](https://github.com/rescript-association/rescript-lang.org/pull/444#discussion_r705067193).

Take a look, and make a call. Happy to go with your preference.

misc_docs/syntax/extension_embed_raw_expression.mdx
kevanstannard: ^^^ I think this is all resolved with your suggestions.

kevanstannard:
Hi @ryyppy, thanks for your suggestions, they look super clean. I believe I've applied all of them.

ryyppy:
> Edit: I didn't apply the link suggestion as they're currently not working for me. Let me know what your preference is here. I'd be happy to take a look at the routing code if you like.

How does this not work? strange. Okay yeah then we need to check what's wrong and fix that first. If you wanna take a stab at this, feel free to do so!

misc_docs/syntax/extension_raw_top_level.mdx
```diff
@@ -0,0 +1,57 @@
+---
```
ryyppy: One last request: Can you rename this file to `extension_raw_top_level_expression.mdx`? Then we are ready to go!

ryyppy:
Perfect, thanks!

PR #446, opened 2021-09-11
Package Index tweaks

kevanstannard:
Hi @ryyppy, I noticed the [discussion about the Package Index recently](https://forum.rescript-lang.org/t/package-index-missing-stuff/2341/16) and took a quick look.

This PR contains a few tweaks that may help, see what you think.

### Main change

The current Fuse.js options for `location` and `distance` favour terms near the start of search fields, but I suspect for the Package Index the location of the term is less important. So I've set the `ignoreLocation` prop to `true` and removed `location` and `distance`.

This small tweak seems to produce reasonable search results. E.g. `zora` and `fast` are returning expected results. Plus a spot check of various terms seems to be working well.

How does it look to you?

### Other changes

[1] Remove `re-schema-form` from `packages_url_resources.json`. It looks like this package is available from npm now. Also fixes a React duplicate key error.

[2] Remove duplicates in the keywords field. It looks like at least one npm package has duplicate keywords, so this fixes a React duplicate key error.

### DIFF
--- a/data/packages_url_resources.json
+++ b/data/packages_url_resources.json
@@ -12,12 +12,5 @@
     "keywords": ["rescript", "typescript"],
     "urlHref": "https://github.com/reason-association/genType",
     "official": true
-  },
-  {
-    "name": "re-schema-form",
-    "description": "Re-schema-form is a meta based render.",
-    "keywords": ["rescript"],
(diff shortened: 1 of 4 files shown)


ryyppy:
LGTM

PR #450, opened 2021-09-29
Typography improvements

bettistein:


styles/_markdown.css
```diff
@@ -14,19 +14,29 @@
 }
 
 .md-ol > .md-li p::before {
-  content: counter(li) ".";
+  content: counter(li) "•";
```
ryyppy, line 17: This is the placeholder character for an ordered list.. should probably be changed back to `1.`, `2.`,.. instead of `1•`, `2•`,...

```suggestion
  content: counter(li) ".";
```

bettistein: Yes, I will do so!

ryyppy:
🎉

PR #457, opened 2021-10-29
Streamline JS method chaining examples

fhammerschmidt:
This [example from the pipe docs](https://rescript-lang.org/docs/manual/latest/pipe#js-method-chaining) is weird, its predecessor uses Js.Array2, while this one relies on extra bindings to map and filter in the prelude.

I deleted the superfluous bindings and used Js.Array2 as well here, instead.


ryyppy:
good change. Thanks!

PR #462, opened 2021-11-25
Add example of importing a react component from JS

traviskaufman:
There is currently no documentation showing how to import an external react component into ReScript. This commit adds one, adopting an example from https://forum.rescript-lang.org/t/use-reactjs-components-in-rescript-react-project/1992

## Example

http://localhost:3000/docs/manual/latest/import-from-export-to-js#import-a-react-component

image

### DIFF
--- a/data/sidebar_react_latest.json
+++ b/data/sidebar_react_latest.json
@@ -1,9 +1,5 @@
 {
-  "Overview": [
-    "introduction",
-    "installation",
-    "migrate-from-reason-react"
-  ],
+  "Overview": ["introduction", "installation", "migrate-from-reason-react"],
   "Main Concepts": [
     "elements-and-jsx",
     "rendering-elements",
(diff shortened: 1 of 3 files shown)


traviskaufman:
Just wondering if anybody's gotten a chance to look at this? I know you're all busy and I appreciate you taking the time to take a look ❤️

ryyppy:
This example should probably go into the rescript-react docs section.
This could even be more prominent by introducing a new `Interop` section in the rescript-react navigation!

traviskaufman:
No worries at all :) Done! 
image

Lmk what you think.

ryyppy:
Wonderful! thanks for that.

I will be damned, but after looking at the test deployment, I don't know why I suggested `JavaScript Interop`, since this terminology is completely inconsistent with the rest of the docs.

Maybe good for discussion: What if we called it `Import / Export ReactJS`? That way we could also give some example on how to consume a ReScript/React component on the JS side?

So instead of `javascript-interop`, call it `import-export-reactjs` and give another example for the export part?

Sorry if I am so nitpicky with that one. Changes on the filesystem level are usually breaking changes, and this needs to be right on the first try :)

fhammerschmidt:
Superseded by #886

PR #466, opened 2021-12-17
Fix typo in let-binding.mdx: %private -> %%private

greyblake:
There seem to be a typo in the documentation.
It should be `%%private` instead of `%private`.

`%%private` is what in the code example and what actually compiles: https://rescript-lang.org/try?code=LYewJgrgNgpgBAQTgXjgbwFBzgUhwBwCcBLANwEMAXGAClkrnJTgGYBKLOeuAI2YBYMAXwxA


ryyppy:
Thanks!

PR #470, opened 2022-01-04
fix(playground): issue with gutter markers not being cleared correctly

liby:
## Screen recording

https://user-images.githubusercontent.com/38807139/148094453-d2ce215f-5be2-484a-9533-262c9789bcf6.mp4

## Description

The issue is mainly due to the fact that gutter markers is not cleared correctly when the `doc.clearGutter` method of CodeMirror is called

## Why not create issue or PR to CodeMirror

1. I'm not sure if ReScript Playground is using the `setGutterMarker` method correctly here
2. It appears that CodeMirror does not provide the feature to provide gutter markers for the columns
    - As you can see from the screen recording, the error marker at the bottom also does not pinpoint the error column

## Other

I'm just a newbie in ReScript, so I'm not sure if the changes here are appropriate.

### DIFF
--- a/src/components/CodeMirror.mjs
+++ b/src/components/CodeMirror.mjs
@@ -2,11 +2,14 @@
 
 import * as Curry from "rescript/lib/es6/curry.js";
 import * as React from "react";
+import * as Belt_Id from "rescript/lib/es6/belt_Id.js";
 import * as Belt_Int from "rescript/lib/es6/belt_Int.js";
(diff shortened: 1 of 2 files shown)


ryyppy:
Is it just me or was the not formatted via `rescript format`?

liby:
> Is it just me or was the file not formatted via rescript format?

I've already executed `rescript format` on my end

> Do you have any reliable way to trigger the wrong behavior?

This bug occurs if multiple code errors are throw on the same line.

PR #471, opened 2022-01-07
Fix format button hover color in playground

vamsikri-hash:
Currently text on format button in playground is not visible when we hover on it. Updated format button hover color to use `text-white-80-tr`.


ryyppy:
Nice! Thank you.

PR #476, opened 2022-01-26
Clarify note in pattern-matching-destructuring.mdx

parke:
Clarify the note regarding `when` clauses versus `if` clauses regarding ReScript 9.0.

I found the old wording to be confusing, so I tried to improve it.  Please let me know if you want an explanation of what exactly I found to be confusing.


ryyppy:
Yeah that might be easier to understand for non-native speakers / language devs. Thanks!

PR #480, opened 2022-02-02
add Arizon logo to users of ReScript

zth:
At Arizon we use ReScript for most things - the startups we build, for client work, and for internal tools.

Here's how the logo looks when added:
image

Thank you for having a look!


ryyppy:
manually rebased into master!

PR #483, opened 2022-02-03
Doc: Beautify pipe code example slightly more

redbar0n:


ryyppy:
We intentionally didn't use any `open Js.Array2` to keep topics isolated

fhammerschmidt:
While that may be off-putting to JS devs at first sight, usually we don't open the module before a pipe-chain since sometimes you have e.g. `myArray->Array.map(doWhatever)->Map.String.fromArray->Map.String....` and then you immediately see that it is now a Map.String on which the operation continues.

Closing this now.

PR #486, opened 2022-02-03
add ohne-makler.net logo

hendi:
This adds the logo for ohne-makler.net, where we're using ReScript for our messaging app which @ryyppy has kindly ported to ReScript from Reason.

image


ryyppy:
Done 🎉

PR #490, opened 2022-02-03
Adding BeOp

remitbri:
Our front-end repo is actually, according to GitHub, 98.7% ReScript! :) (786 modules as today)


ryyppy:
Thanks!

PR #491, opened 2022-02-04
installation page updated

coleea:
hello.

First, this is my first PR ever. So some procedural error might exist.

I use M1 mac mini. Yesterday I tried to install rescript using `npm i -g rescript`. but failed because of errors.

I searched for internet and I found many people failed to install rescript on `M1 mac device`.

As I know, the best solution for this is downgrading npm version to 6.14. It seems like that npm version `8.x` triggers installation error.

My conclusion is that the installation guide page `https://rescript-lang.org/docs/manual/latest/installation` need to be updated about M1 installation fail scenarios and solutions such as installing `npm@6.14`. If not, many users even not try to rescript at all.

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,0 +1,1 @@
+[0204/141126.005:ERROR:registration_protocol_win.cc(102)] CreateFile: 지정된 파일을 찾을 수 없습니다. (0x2)
--- a/pages/docs/manual/latest/installation.mdx
+++ b/pages/docs/manual/latest/installation.mdx
@@ -30,9 +30,14 @@ During development, instead of running `npm run build` each time to compile, use
 If you already have a JavaScript project into which you'd like to add ReScript:
 
 - Install ReScript locally as a [devDependency](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file):
-  ```sh
+  ```shhttps://github.com/coleea/rescript-lang.org
   npm install rescript --save-dev 
   ```
+  
+  If you use M1-based Mac Device, error might be occur. If so, try installing `npm@6.14` via `npm i -g npm@6.14`, then try again `npm install rescript --save-dev`.
+   
+
+
 - Create a ReScript build configuration at the root:
   ```json
   {


debug.log
```diff
@@ -0,0 +1 @@
+[0204/141126.005:ERROR:registration_protocol_win.cc(102)] CreateFile: ����� ����� ã� �� ���ϴ�. (0x2)
```
ryyppy, line 1: This file should not be committed

pages/docs/manual/latest/installation.mdx
```diff
@@ -30,9 +30,14 @@ During development, instead of running `npm run build` each time to compile, use
 If you already have a JavaScript project into which you'd like to add ReScript:
 
 - Install ReScript locally as a [devDependency](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file):
-  ```sh
+  ```shhttps://github.com/coleea/rescript-lang.org
```
ryyppy, line 33:
```suggestion
  ```sh
```

pages/docs/manual/latest/installation.mdx
```diff
@@ -30,9 +30,14 @@ During development, instead of running `npm run build` each time to compile, use
...
   npm install rescript --save-dev 
   ```
+  
+  If you use M1-based Mac Device, error might be occur. If so, try installing `npm@6.14` via `npm i -g npm@6.14`, then try again `npm install rescript --save-dev`.
```
ryyppy, line 37: This note should probably be part of the `Prerequisite` headline on the very top. This important m1 note gets missed otherwise?

coleea: Yes. I also think that Prerequisite section is better suitable.
Also, I found the case that installing npm@6.14  could not solve the error. In this case, the problem solved when using rosetta. So I suggest that some link for Rosetta tutorial might helpful for people. for example
https://derflounder.wordpress.com/2020/11/17/installing-rosetta-2-on-apple-silicon-macs/

by the way, should I do pull-request again after re-writing the document file or Is it ok to do nothing ? I ask for your understanding about git-novice (maybe silly question).
thank you

Zeta611: @coleea If you fix the issues and `git push` again, it will be automatically reflected in this PR page.

ryyppy:
We now ship M1 binaries, and I think this PR is now kinda obsolete. Can you verify if you are still running into issues with the current version of the docs? If yes, feel free to reopen.

PR #502, opened 2022-03-14
Add DevITjobs as a user

MariaAbadi:


ryyppy:
image

Can you modify the SVG in a way that it uses `#979AAD` as the fill color instead of blue? Otherwise it will not be consistent with the rest of the logos.

ryyppy:
perfect! thanks 👍

PR #511, opened 2022-04-27
♻️ update the copyright year.

shaunchander:
I noticed the copyright year says "2021", updated it to say "2022".


src/common/OurUsers.mjs
```diff
@@ -121,6 +121,12 @@ var companies = [
     url: "https://camelohq.com",
     path: "/static/lp/camelo.svg",
     [Symbol.for("name")]: "Logo"
+  },
+  {
+    name: "DevIT Jobs",
```
ryyppy, line 126: was this an intentional change? if so, could you factor that into its own PR?

shaunchander: No this change was unintentional, not sure how that happened though 😅.

robinweser:
Why don't we just use `new Date().getFullYear()` instead? Then we don't have to change that every year :P

shaunchander:
> Why don't we just use `new Date().getFullYear()` instead? Then we don't have to change that every year :P

I'll add this in.

Grubba27:
I saw this issue and i've done the function... https://github.com/rescript-association/rescript-lang.org/pull/519/files, if it is better I can just place it here and shaunchander can use in this pr :)

ryyppy:
Fixed in #521, due to it not having unwanted file diffs in there. Will check what's up with the `package-lock.json` changes.

Thanks for contributing!

PR #516, opened 2022-05-11
Add Reka.market logo

PolyDevil:
We use rescript for telegram bot, our design system and for i18n of main nextJS app.


ryyppy:
thanks! the logo seems to be too close to the other logos, but this is something we need to fix with #524

PR #518, opened 2022-05-14
Corrected React Style Type reference

### LINKED ISSUE #517: The module or file React.Style can't be found
Following along in the docs here: https://rescript-lang.org/docs/react/latest/styling#css-in-js
led to this error: `The module or file React.Style can't be found`

From the docs:
```rescript
@module("@emotion/css") external css: React.Style.t => string = "css"
```

But it looks like it should be:
```rescript
@module("@emotion/css") external css: ReactDOM.Style.t => string = "css"
```

bobfp:
Fixes #517


ryyppy:
thanks!

PR #519, opened 2022-05-16
Fix(footer): Adjusted to always get current year

Grubba27:
I was scrolling thought the lang page and I saw the year was not in the current year.
Added a function that places the current year as copyright.
Don't know hot it works in the US... 
In pages from my country I've added this function to always place the current year in this section of the footer.
Also It could be a small adjust to place year of launch - <now>
If there is anything that needs to be changed I'm here to help :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "rescript-lang.org",
       "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


ryyppy:
fixed with #521

ryyppy:
Thanks for taking this on though. This PR had a diff in the package-lock.json, so I merged the equivalent PR mentioned above.

PR #520, opened 2022-05-16
add @live/@dead/@raises/@doesNotRaise decorators from reanalyze to syntax lookup

zth:
cc @ryyppy @cristianoc 

Very open to feedback here. I wanted to add these two to the official documentation (even though they're strictly from `reanalyze`) because I'm using them quite a lot in my code bases, and I've noticed people try to look them up, but struggling to find them. I personally thing `reanalyze` is so good we should push it more, therefore adding these here.

### DIFF
--- a/misc_docs/syntax/decorator_dead.mdx
+++ b/misc_docs/syntax/decorator_dead.mdx
@@ -1,0 +1,15 @@
+---
+id: "dead-decorator"
+keywords: ["dead", "decorator"]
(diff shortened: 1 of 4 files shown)


zth:
@ryyppy sounds good, will fix tomorrow.

zth:
@ryyppy mind taking another look?

ryyppy:
thanks!

PR #533, opened 2022-06-15
fix: use flex to position items side-by-side

davesnx:
Fixes a small position issue in Markdown rendering for list items:

## Before
Screenshot 2022-06-15 at 11 35 59

## After
Screenshot 2022-06-15 at 11 36 22


ryyppy:
thanks!

PR #534, opened 2022-06-24
add DocSearch v3

### LINKED ISSUE #245: Search results should display Git tag/docs version
## Problem

A search of `switch` will result in multiple results for seemingly the same thing, but in fact the results are for multiple versions of the rescript docs, this results in having the docs version changed when clicking on the wrong link and not easily knowing you did switch. The desired link is only known if you know the version of the docs you are on AND look at the URL listed by the browser (usually bottom-left of window) on hover.

This issue is amplified by the fact the older docs do not have Rescript syntax and for those who do not understand how or that they need to change the version can be cumbersome and lead to issues trying to understand why following the docs is leading to errors (EG: Rescript using `array<>` syntax over `array()`)

rescript-docs-search-bug

## Expected

The search should only list results for URLs of the currently selected version of the docs, or also specify the verison clearly in the results.

This may not be easily resolved :shrug: 

PS: Rescript is awesome, thanks for all your work.

### LINKED ISSUE #465: Searching for "Array" results in links that are difficult to distinguish
Capture d’écran, le 2021-12-12 à 15 39 35

After clicking on four of the links it also looks like two of them were duplicates.

### LINKED ISSUE #498: Infinite loop from blocked request
Hi, when I try to access rescript-lang.org/ the page stays blank and my browser makes an infinite loop of requests.
Here's a screenshot of how it looks.
Screen Shot 2022-03-03 at 10 45 44 PM

It seems like one request is being blocked (I'm using Firefox with uBlock Origin with 3rd party scripts blocked by default), and that seems to cause the issue.  It would be great if this could be handled more gracefully.

Thanks!

### LINKED ISSUE #528: Some workaround for the blocking of jsdelivr.
The official website has been blocked for days in china. Just because of the cdn `jsdelivr` is blocking may be blocked forever.
I found that just import the library docsearch from cdn into local in this project, the docs and api pages can be access normally(this is enough for us).
image

aspeddro:
## Changes

- Add bindings to `@docsearch/react` v3

## UI Changes

- Search icon is now visible in any width. Before, it was not visible between 599px and 766px
- Remove the search icon from the mobile menu and keep it always visible. This is the behavior on sites like [reactnative.dev](https://reactnative.dev/), [tailwindcss](https://tailwindcss.com/), and [reactjs](https://beta.reactjs.org/)

- [x] Update [record extractor](https://docsearch.algolia.com/docs/record-extractor)
  - See [indexing content for faceting](https://docsearch.algolia.com/docs/record-extractor#indexing-content-for-faceting) #245

Fix #498
Fix #528
Fix #245
Fix #465


ryyppy:
Thanks for the PR.

So generally this looks good, although I am not entirely convinced about algolia's way of design here and at some point I wanted to have something custom tailored. Anyways, the only nitpick: This doesn't really fix #465 since we'd need to fix the Algolia crawling and UI mapping first (as stated in your TODO section). The new experience is even more confusing, since it will render empty rows.

image

Still wondering if we should merge this and then follow up with the relevant index changes.

aspeddro:
> So generally this looks good, although I am not entirely convinced about algolia's way of design here and at some point I wanted to have something custom tailored.

Okay, I'll refactor it.

> This doesn't really fix https://github.com/rescript-association/rescript-lang.org/issues/465 since we'd need to fix the Algolia crawling and UI mapping first (as stated in your TODO section). 

Sure, Fixed.


> The new experience is even more confusing, since it will render empty rows.


This is due to the new version of the extractor, which must be updated.

aspeddro:
> Still wondering if we should merge this and then follow up with the relevant index changes.

Is it possible to create a new extractor so that we can test it?

ryyppy:
I have access to the new editor experience on the algolia website. Will need to update the api keys and then try to work my way around the crawling configuration. Will let you know as soon as I got something going.

aspeddro:
Now the design is more personalized. I changed it to the `DocSearchModal` component.

aspeddro:
https://github.com/rescript-association/rescript-lang.org/pull/534/commits/2e22535d68710a1b265ca0e8cededaa32a017b7b add an alternative to not rendering empty lines.

image

aspeddro:
I don't think we need to change the crawler. Git tags are displayed.

image

ryyppy:
can we only show results for the latest version? We may make this dependent on the current version that is being browsed, but for the start, only latest would be great.

aspeddro:
Ready for review

ryyppy (approved):

PR #535, opened 2022-06-28
[Fix] Correct typo "re" to "res" in GenType

Zeta611:
Hi, this is a small typo fix that corrects a broken link (from `.re` to `.res`).


cristianoc:
Thanks.

PR #536, opened 2022-07-05
add space between logos

davesnx:
### Before

Screenshot 2022-07-05 at 15 34 55

### After
Screenshot 2022-07-05 at 15 28 28


ryyppy:
thanks!

PR #555, opened 2022-09-15
Update build schema to the one of 10.0.

cristianoc:


ryyppy (approved):
LGTM. Only issue is that the `build_schema` isn't really versioned. So some v9 users may believe that certain v10 settings are available in their compiler version.

cristianoc:
Removed change of compiler --  let's see if it makes vercel happy.

Issue #556, opened 2022-09-15
Improve documentation of Integers


johanalkstal:
I find the current description of why you need to be careful when working with JavaScript numbers a bit unclear on the "why" you should be careful. I think it could be explained a little bit more, without expanding the documentation much.

I'd like to suggest the following change,

`-`
```
**Careful when you bind to JavaScript numbers**! Long ones might be truncated. Bind JS number (especially Dates) as **float** instead.
```

`+`
```
**Be careful when you bind to JavaScript numbers!** Since ReScript integers have a much smaller range than JavaScript numbers, data might get lost when dealing with large numbers. In those cases it’s much safer to bind the numbers as **float**. Be extra mindful of this when binding to JavaScript Dates and their epoch time.
```

What do you think?

cristianoc:
Great, would you create a PR?

PR #557, opened 2022-09-15
Update primitive-types.mdx

johanalkstal:
The description of Integers was not as clear on why you should be careful when binding to numbers as it could be. Tried to describe a bit more why you should be careful.

https://github.com/rescript-association/rescript-lang.org/issues/556


cristianoc (approved):
Thanks!

Issue #552, opened 2022-09-08
Create a v10.0 version of docs.


cristianoc:
Here are the verbatim release notes: for each item, need to evaluate whether some doc update is required.

# 10.0.1

#### :bug: Bug Fix

- [x]  Fix issue where watch mode would give an error on Windows https://github.com/rescript-lang/rescript-compiler/pull/5621

# 10.0.0

**Compiler**

#### :boom: Breaking Change

- [x] `bsconfig.json` does not support `// line` comments anymore.
  - Example: `"suffix": ".bs.js" // determine the suffix`
  - Fix: remove the comment and use standard json.
- [x] Changed return type of `Js.String.match_` as it was wrong. [#5070](https://github.com/rescript-lang/rescript-compiler/pull/5070)
  - Example: any use of `Js.String.match_` and `Js.String2.match_`
  - Fix: follow the type errors

#### :rocket: New Feature

- [x] New records with optional fields e.g. `type opt = {x: int, y?: string}` were added as an experimental feature [#5423](https://github.com/rescript-lang/rescript-compiler/pull/5423) [#5452](https://github.com/rescript-lang/rescript-compiler/issues/5452) [New Syntax](https://github.com/rescript-lang/syntax/pull/589/files)
- [x] Add support for `@new @variadic` (see https://github.com/rescript-lang/rescript-compiler/pull/5364)

#### :bug: Bug Fix

- [x] Classify bigint correctly [#5351](https://github.com/rescript-lang/rescript-compiler/pull/5351)
- [x] Fixed crash in `rescript build` on Windows [#5516](https://github.com/rescript-lang/rescript-compiler/pull/5516)
- [x] Fixed `rescript init` command not working [#5526](https://github.com/rescript-lang/rescript-compiler/pull/5526)
- [x] Fix issue with compiler log not terminated that causes problems with editor extension not clearing issues when fixed [#5545](https://github.com/rescript-lang/rescript-compiler/issues/5545)

##### :nail_care: Polish

- [x] Changed Linux build to depend on GLIBC 2.28 again for compatibility with Debian 10.

- [x] Proper M1 support (CI now supports M1 native builds)

**Syntax**

#### :boom: Breaking Change

- [x] `@bs.send.pipe` is now removed. Earlier it was deprecated.
- [x] Missing labels in function application is now an error (https://forum.rescript-lang.org/t/ann-more-strict-checks-in-missed-labels/2117).
[...]

PR #560, opened 2022-09-20
Adjust version for 10.0 docs

cristianoc:
Will use "latest" just as `9.1`.
New features/changes can be added by mentioning they're new in 10.0.


cristianoc:
https://github.com/rescript-association/rescript-lang.org/issues/552

PR #566, opened 2022-09-26
docs for optional record fields

zth:
A bit crude, feedback welcome.

### DIFF
--- a/pages/docs/manual/latest/record.mdx
+++ b/pages/docs/manual/latest/record.mdx
@@ -20,7 +20,7 @@ A record needs a mandatory type declaration:
 ```res prelude
 type person = {
   age: int,
-  name: string
+  name: string,
 }
 ```
 ```js
@@ -28,6 +28,28 @@ type person = {
 ```
 
 </CodeTab>
+
+The value of the record fields can be declared as `option`, just like you'd expect. In addition, the presence of the record fields themselves can also be specified as optional, by putting a `?` after the field name:
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res prelude
+type person = {
+  age: int,
+  name?: string
+}
+```
+```js
+// Empty output
+```
+
+</CodeTab>
+
+This has 2 effects, both of which we'll dive deeper into later:
+1. You don't need to specify the field when creating the record
+2. Its type is automatically option, and the value defaults to `None` when the field is not specified
+
+The resulting type of the `name` field above will therefore be `option<string>`.
 
 ## Creation
 
@@ -90,6 +112,41 @@ var me2 = {
 
 Either of the above 3 says "this record's definition is found in the School file". The first one, the regular type annotation, is preferred.
 
+### Optional Fields
+For records with optional fields, just set the fields you're interested in:
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res prelude
+type person = {
+  age: int,
+  name?: string
+}
+
+let me = {
+  age: 5,
+  name: "Big ReScript"
+}
+
+let friend = {
+  age: 7
+}
+```
+```js
+var me = {
+  age: 5,
+  name: "Big ReScript"
+};
+
+var friend = {
+  age: 7
+};
+```
+
+</CodeTab>
+
+Notice how the fields not set will also not end up in the JavaScript output.
+
 ## Access
 
 Use the familiar dot notation:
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/record.mdx
```diff
@@ -20,7 +20,23 @@ A record needs a mandatory type declaration:
...
+  name: string,
+}
+```
+```js
+// Empty output
+```
+
+</CodeTab>
+
+The value of the record fields can be declared as `option`, just like you'd expect. In addition, the presence of the record fields themselves can also be specified as optional, by putting a `?` after the field name:
```
cristianoc, line 32: I would drop the mention of `option`, or explain the difference.
Also, it should say upfront that this is from version `10.0`. I see that mentioned below, but I guess it should be mentioned every time? Definitely the first time.

pages/docs/manual/latest/record.mdx
```diff
@@ -90,6 +112,41 @@ var me2 = {
...
+};
+
+var friend = {
+  age: 7
+};
+```
+
+</CodeTab>
+
+Notice how the fields not set will also not end up in the JavaScript output.
```
cristianoc, line 148: Double negation: re-phrase?

pages/docs/manual/latest/record.mdx
cristianoc: I guess 2 possilble patterns are:

1 Optional Fields (since rescript 10.0) ...
for every piece of doc

2 move all the optional fields paragraphs in one place

which one I guess depends on which learning journey one wants to optimize
someone who learns about records for the first time probably gets a gentler introduction with the latter choice
someone who's looking for updates specifically, might benefit from having all the update docs in one place

pages/docs/manual/latest/record.mdx
```diff
@@ -125,6 +182,44 @@ var meNextYear = {
 
 **Note**: spread cannot add new fields to the record value, as a record's shape is fixed by its type.
 
+### Optional Fields
+Updating an optional field in the scenario above works just like any other field provided you want to set a new value. However, if you want to _remove_ the value of that field, you'll need to use a special syntax. Note that to remove a field here means to set its value to `None`. It looks like this:
```
cristianoc, line 186: other field, provided y

pages/docs/manual/latest/record.mdx
```diff
@@ -125,6 +182,44 @@ var meNextYear = {
...
+}
+
+let me = {
+  age: 123,
+  name: "Hello"
+}
+
+let withoutName = {
+  ...me,
+  name: ?None
```
cristianoc, line 203: Somewhere the info should be conveyed that this is not specific to `None` and any expression of option type can go there.

pages/docs/manual/latest/record.mdx
```diff
@@ -189,6 +285,112 @@ getAge(me) // type error!
...
 
+### Optional Fields in Records Can Be Useful for Bindings
+Many JavaScript APIs tend to have large configuration objects that can be a bit annoying to model as records, since you previously always needed to specify all record fields when creating a record. 
+
+Optional record fields, introduced in [`v10`](/blog/release-10-0-0#experimental-optional-record-fields), is intended to help with this. Optional fields will let you avoid having to specify all fields, and let you just specify the one's you care about. A significant improvement in ergonomics for bindings and other APIs with for example large configuration objects.
+
+### Pattern Matching on Optional Fields
+[Pattern matching](pattern-matching-destructuring), one of ReScript's most important features, has two caveats when you deal with optional fields.
+
+When matching on the field directly, the value of the field is always treated as `option`. Example:
```
cristianoc, line 296: The example is illustrative, but it does not really matter that the value you're matching comes from field access.
Perhaps change the phrasing so that this example is the old way, and the next 2 examples are the new way?

cristianoc (commented):
Left a few comments.
Take whatever makes sense from them.

pages/docs/manual/latest/record.mdx
```diff
@@ -157,9 +157,247 @@ Fields not marked with `mutable` in the type declaration cannot be mutated.
...
+  age: 7
+};
+```
+
+</CodeTab>
+
+This has consequences for pattern matching, which we'll expand a bit on soon.
+
+## Immutable Update
+Updating an optional field via an immutable update above let's you set that field value without needing to care whether it's optional or not.
```
cristianoc, line 215: lets

cristianoc (approved):

PR #567, opened 2022-09-27
Some mention of doc comments.

cristianoc:
https://github.com/rescript-association/rescript-lang.org/issues/552

### DIFF
--- a/pages/docs/manual/latest/overview.mdx
+++ b/pages/docs/manual/latest/overview.mdx
@@ -21,8 +21,8 @@ canonical: "/docs/manual/latest/overview"
 | -------------------- | -------------------------------- |
 | `// Line comment`    | Same                             |
 | `/* Comment */`      | Same                             |
-| `/** Doc Comment */` | `/** Doc Comment Attached */`    |
-| No distinction       | `/*** Standalone Doc Comment */` |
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/overview.mdx
```diff
@@ -17,10 +17,12 @@ canonical: "/docs/manual/latest/overview"
...
 
-| JavaScript        | ReScript    |
-| ----------------- | ----------- |
-| `/* Comment */`   | Same        |
-| `// Line comment` | Same        |
+| JavaScript           | ReScript                         |
+| -------------------- | -------------------------------- |
+| `// Line comment`    | Same                             |
+| `/* Comment */`      | Same                             |
+| `/** Doc Comment */` | `/** Doc Comment Attached */`    |
```
zth, line 24: Might be worth clarifying what "attached" means. It reads like it's different to JS (since it's not marked as "Same" in the ReScript column).

zth (commented):
Just a question!

pages/docs/manual/latest/overview.mdx
cristianoc: Better now?

zth (approved):

PR #574, opened 2022-10-10
Use the res prefix for convenience npm scripts

DZakh:
> There are some unrelated autoformatting changes. I thought they won't harm, so decided to keep them.


cristianoc (approved):
Thanks

PR #575, opened 2022-10-11
Update installation guilde: Rename convenience npm script from res:start to res:dev

DZakh:
The logic behind the PR is that recently the `start` script often started being used in the context of running a built application. While scripts for development are usually called `dev` or `watch`.

- https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-start

### DIFF
--- a/pages/docs/manual/latest/installation.mdx
+++ b/pages/docs/manual/latest/installation.mdx
@@ -58,7 +58,7 @@ If you already have a JavaScript project into which you'd like to add ReScript:
   ```json
   "scripts": {
     "res:build": "rescript",
-    "res:start": "rescript build -w"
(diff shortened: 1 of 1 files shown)


DZakh:
> For the sake of semantics, I'd rather have res:watch than res:dev, since the latter would look similar to npm run dev which is very common in e.g. a NextJS project.

That's actually the point. So when you start developing a nextjs application, you'd run commands with similar names/meanings:
- `npm run relay:dev` + `npm run res:dev` + `npm run next:dev`
- `npm run res:dev` + `npm run next:dev`
- `npm run res:dev` + `npm run dev`
- `npm run dev` where `"dev": "npm-run-all res:build --parallel --print-label res:dev next:dev"`

PR #579, opened 2022-10-18
correct docs for char in rescript about unicode

bobzhang:


cristianoc:
Would you add: since compiler 10.0 somewhere.
As the docs are shared.

cristianoc (approved):
Thanks!

PR #582, opened 2022-10-23
Removing extra % from docs

Grubba27:
Found this typo today while doing some interop


cristianoc:
See below: "While `%%raw` lets you embed top-level raw JS code, `%raw` lets you embed expression-level JS code:"

PR #584, opened 2022-10-28
fix: An building error on windows platform which caused by esmodule _…

JinweiOS:
When I join the work about traslating rescript-lang site to chinese [here](https://github.com/butterunderflow/rescript-lang.org-chinese-translation), I have found an error of building the site project. so I added some code to support unix-like and win32 system at the same time.


ryyppy (approved):
lgtm!

PR #590, opened 2022-11-05
Add documentation for async/await.

cristianoc:


pages/docs/manual/latest/function.mdx
```diff
@@ -415,6 +415,78 @@ If you write down the uncurried function's type, you'll add a dot there as well.
 
 **This feature seems trivial**, but is actually one of our most important features, as a primarily functional language. We encourage you to use it if you'd like to remove any mention of `Curry` runtime in the JS output.
 
+## Async/Await (from version v10.1)
+
+Just as in JS, an async function can be declared by adding `async` before the definition, and `await` can be used in the body of such functions.
+The outpul looks like idiomatic JS:
```
zth, line 421: Typo outpul

pages/docs/manual/latest/function.mdx
```diff
@@ -415,6 +415,78 @@ If you write down the uncurried function's type, you'll add a dot there as well.
...
+```
+```js
+async function greetUser(userId) {
+  var name = await getUserName(userId);
+  return "Hello " + name + "!";
+}
+```
+</CodeTab>
+
+The return type of `getUser` is inferred to be `promise<string>`, where type `promise` is now a built-in of the language.
```
zth, line 441: Not sure we need to mention that the promise type is now built in, might just confuse and it's an implementation detail that most users don't need to know about.

pages/docs/manual/latest/function.mdx
```diff
@@ -415,6 +415,78 @@ If you write down the uncurried function's type, you'll add a dot there as well.
...
+let someAsyncFn = async () => {
+  switch await somethingThatMightThrow() {
+  | data => Some(data)
+  | exception JsError(_) => None
+  | exception SomeReScriptException => None
+  }
+}
+```
+```js
+var SomeReScriptException = /* @__PURE__ */Caml_exceptions.create("JsxV4.SomeReScriptException");
```
zth, line 466: JsxV4 might be confusing here, but not sure we need to care tbh.

zth (approved):
Great!

src/components/Navigation.mjs
```diff
@@ -602,13 +602,13 @@ function Navigation(Props) {
                             }, React.createElement("div", {
                                   className: "hidden sm:block mr-6"
                                 }, React.createElement(DocSearch.make, {})), React.createElement("a", {
-                                  className: "mr-5 no-underline block text-inherit hover:cursor-pointer hover:text-fire-30 text-gray-40 mb-px",
+                                  className: "mr-5 " + link,
```
cristianoc, line 605: only inlining difference

zth (approved):
Awesome! 👍

ryyppy:
I guess there also need to be changes for the Async & Promise section as well?

Issue #589, opened 2022-11-04
JSX page needs overhaul


fhammerschmidt:
https://rescript-lang.org/docs/manual/latest/jsx

* it still mentions ReasonReact
* it even links to the old ReasonReact docs
* children spread is not necessary anymore 
* the "JS Output" is often not the real final JS output

Overall I am also not sure if the desugared / ppxed snippets should be there anyway.

Of course now with JSX4 landing, it will be adapted anyway, those were just the things that stand out like a sore thumb to me.

mattdamon108:
IMHO, I couldn’t find the fine reason to state the desugared expressions by ppx.
I think it is better to explain what `@react.component` is doing for.

cristianoc:
The whole `@JSX` description can go as `@JSX` has now become an implementation detail that can just disappear now that the parser is vertically integrated in the compiler.

mattdamon108:
> I think it is better to explain what `@react.component` is doing for.

Nope. I guess `@react.component` is not necessary to be explained on the JSX page. It is more like to be stated on the ReScriptReact page.

PR #591, opened 2022-11-05
Update JSX page

mattdamon108:
Follow up the issue https://github.com/rescript-association/rescript-lang.org/issues/589

### DIFF
--- a/pages/docs/manual/latest/jsx.mdx
+++ b/pages/docs/manual/latest/jsx.mdx
@@ -162,12 +162,12 @@ React.createElement(MyComponent, {
 ## Departures From JS JSX
 
 - Attributes and children don't mandate `{}`, but we show them anyway for ease of learning. Once you format your file, some of them go away and some turn into parentheses.
-- The spread props is supported, but there are some restrictions.
+- Props spread is supported, but there are some restrictions (see below).
 - Punning!
 
-### Spread props
+### Spread Props (from v10.1)
 
-The JSX spread props is supported.
+JSX props spread is supported now, but in a stricter way than in JS.
 
 <CodeTab labels={["ReScript", "JS Output"]}>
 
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/jsx.mdx
```diff
@@ -108,153 +106,132 @@ React.createElement(React.Fragment, undefined, child1, child2);
...
+React.createElement(Comp, {
+  a: "a",
+  b: "b"
+});
 ```
 
 </CodeTab>
 
-This is extra useful in the cases where you are handled `myChild` that is already a list of things, and want to forward that without wrapping it an extra time (which would be a type error) \*. It also allows you to pass arbitrary data structures at `children` position (remember, JSX `children` is really just a totally normal prop):
+The multiple spreads are not allowed:
```
fhammerschmidt, line 186: "Multiple spreads are not allowed:" 
(ditch the `The`)

pages/docs/manual/latest/jsx.mdx
```diff
@@ -108,153 +106,132 @@ React.createElement(React.Fragment, undefined, child1, child2);
...
 ```
 
 </CodeTab>
 
-## Usage
-
-See [ReasonReact JSX](https://reasonml.github.io/reason-react/docs/en/jsx) for an example application of JSX, which transforms the above calls into a ReasonReact-specific call.
-
-Here's a JSX tag that shows most of the features.
+The spread must be the first position followed by other props:
```
fhammerschmidt, line 196: "The spread must be **at** the first position, followed by other props:"

fhammerschmidt (changes requested):

mattdamon108:
@fhammerschmidt Thanks! I've fixed all.

fhammerschmidt (approved):

ryyppy:
looks good. shall we merge?

PR #597, opened 2022-11-09
Update genType docs after the move inside the compiler.

cristianoc:


pages/docs/gentype/latest/getting-started.mdx
```diff
@@ -50,7 +41,7 @@ structure:
 ```js
   //...
   "gentypeconfig": {
-    "language": "typescript" | "flow" | "untyped",
+    "language": "typescript",
```
zth, line 44: Not for now, but should we remove this configuration option all together now that it just supports one? Or am I misunderstanding?

zth (approved):
Awesome!

pages/docs/gentype/latest/getting-started.mdx
cristianoc: If you're using compiler 10.0 or earlier, this is still required. I'll say that explicitly.

PR #602, opened 2022-11-14
fix: syntax lookup module keyword

aspeddro:
The language constructor `module` show content of `__MODULE__` special value.


cristianoc:
Thanks for the fix.

PR #604, opened 2022-11-14
Fix simple calculation mistake in Belt.Array.map code example

Yisuk:
Hello 😄 
The current code example of `Belt.Array.map` has simple calculation error. 
Screen Shot 2022-11-14 at 6 32 00 PM

So I changed `f` from `(x) => x + 1` to `(x) => x + 2`.
Screen Shot 2022-11-14 at 6 38 23 PM


cristianoc:
Thanks!

PR #620, opened 2022-12-08
CI: format check

aspeddro:
rescript v10.1.0 has a lot of change in formatting

This change is to ensure that the formatting is always up to date.


cristianoc:
Thanks

aspeddro:
I just found out that just run `npx rescript format -check -all`

PR #622, opened 2022-12-09
Remove checked in `.mjs` build artifacts

### LINKED ISSUE #616: Remove checked in .mjs build artifacts and add them to gitignore
I think we reached a point where we can rely on the compiled JS output.

aspeddro:
Close #616


ryyppy:
Amazing. Thanks!

PR #624, opened 2022-12-13
Internationalization for the website

butterunderflow:
Please note, this PR does **not attempt to be merged now**, it is only for review.

I am trying to implement i18n while making as few changes as possible to the existing codebase.

I haven't used the nextjs i18n framework, because it seems difficult to work with the existing codebase, and static page generation would become unsupported.

This PR tries to implement i18n in a kind of brutal way: distribute the content of different locales in different folders, and then make all the layouts "locales-dependent". Finally, add hrefs between pages of different locales. 
For example, I put all Chinese translations into the `pages/zh-CN` folder, and add locale testing before applying the `LatestLayout`. 

Till now the translation and layout of the language manual have been added. And here is a [demo](https://rescript-lang-ppg870b0h-butterunderflow.vercel.app/docs/manual/latest/introduction).

Please let me know if I'm in the right direction, or give me any advice. If this works, I will continue to finish the other parts. Thank you!


ryyppy:
Woah, this is a crazy amount of work. Really appreciate your initiative!

> I haven't used the nextjs i18n framework, because it seems difficult to work with the existing codebase, and static page generation would become unsupported.

I'd be highly in favour in using NextJS builtin functionality. Could you determine what parts would need refactoring so we can make this work? We need to remove any complexities that prevent us from upgrading to the latest NextJS version anyways.

Also what do you mean by "static page generation would become unsupported"? My understanding was that only the routing would be kinda dependent on infrastructure? We pay for Vercel, so unless the Chinese access would be compromised, we may wanna take a stab at solving those issues.

> This PR tries to implement i18n in a kind of brutal way: distribute the content of different locales in different folders, and then make all the layouts "locales-dependent".

Yeah that's pretty brutalistic, and this looks like a ton of work. I will probably repeat this on the forum, but right now we are also in the talks with the Korean community to create translations for the website, but I didn't have the time to summarize / express the next tasks.

One task was to use a translation service like [Crowdin](https://crowdin.com/) to keep track of the EN version and help translators to update new / changed content.

You already translated a lot of the content, so might be worth to connect you and the Korean translator to try putting your translation data to the translation software as well. Crowdin allows us to export the translated markdown files into our Git repository later on.

### Other points regarding the translations:

- Translating API docs: This can't be done for now. In the future we will generate the docs from our source code for this website, so all your translations will be lost / out of date rather quickly. I think translating (API) code should be excluded for now.
- Docs Structure: Our manual and gentype docs need to be merged rather soon (since it will be shipped with the latest `rescript` version). The content will therefore be moved in the `manuals` folder. I'd also want to have a more aligned tone of the `genType` docs, so maybe let's hold out with translations if possible?

butterunderflow:
> I'd be highly in favour in using NextJS builtin functionality. Could you determine what parts would need refactoring so we can make this work? We need to remove any complexities that prevent us from upgrading to the latest NextJS version anyways.

I also prefer the builtin functionality of NextJS. At first, I thought the brutal solution will be"okay", but when I actually tried it, I realized that it required a huge amount of work. I'll try to make another demo with builtin i18n support.

> One task was to use a translation service like [Crowdin](https://crowdin.com/) to keep track of the EN version and help translators to update new / changed content.

> You already translated a lot of the content, so might be worth to connect you and the Korean translator to try putting your translation data to the translation software as well. Crowdin allows us to export the translated markdown files into our Git repository later on. 

The translation was mainly contributed by Chinese ReScript enthusiastic and I am always here to help. 

In addition, for anyone interested, [here](https://github.com/rescript-idea/rescript-lang.org-chinese-translation) is the Chinese fork of the website.

PR #629, opened 2022-12-21
Fix bold effect in introduction doc

JiaeK:
Not sure [this](https://rescript-lang.org/docs/manual/latest/introduction#:~:text=This%20means%20you%20can%20pick%20up%20ReScript%20and%20access%20the%20vast%20JavaScript%20ecosystem%20and%20tooling%20as%20if%20you%27ve%20known%20ReScript%20for%20a%20long%20time!) is intentional but I think it's better when the exclamation mark is also bolded inside the below sentence :) 

<br>

| Before | After | 
|--------|-------|
| bold-before | bold-after |


ryyppy:
well spot. Thanks!

PR #630, opened 2022-12-24
Use latest build schema

pd4d10:
The schema at https://rescript-lang.org/docs/manual/latest/build-configuration-schema is outdated.

In this PR, we fetch and use the latest schema, instead of the one committed previously in the codebase.


ryyppy:
This is pointing to the schema of latest master, which means it may show a schema with changes that are not yet released. I'd be more in favour to pinpoint to a tagged ref url when e.g. navigating to the `latest` configuration schema route.

For older versions, it would pin point to the right version as well.

pd4d10:
> pinpoint to a tagged ref url

Updated

ryyppy:
nice solution. thanks!

PR #645, opened 2023-01-27
Attempt: Upgrade to NextJS version 13

JiaeK:
This PR is an attempt to upgrade this repo’s NextJS version to the latest (v.13). 
The idea stemmed from the talk which initially desired to integrate the internationalization function for rescript-lang.org. I remember the ReScript team wanted the NextJS upgrade to happen with the i8n, so we (with @namenu & @mununki) thought why don’t we help? 

Up to NextJS version 12, the upgrade was decently successful. And to version 13, it’s not completely done since there're a few hydration errors that still exist. Hope this is somewhat helpful to initiate the idea. 

This PR,

- Updates NextJS to v.13 ([ref](https://nextjs.org/docs/upgrading))
- Adjusts `transpileModules` in `next.config.js` ([ref](https://github.com/martpie/next-transpile-modules/releases/tag/the-end))
- Removes `esbuild-loader`, `path`, `fs`, `target`, `future` in `next.config.js`
- Adds `SWC` in `next.config.js`
- Removes `.babelrc` file
- Switches to `@next/mdx`
- Removes child `<a>` tags of `<link>`


ryyppy:
Just as a side-note, the frontmatter parsing seems to be broken as well:
image

But this is a good start!

pages/docs/react/latest/migrate-from-v3.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: Migrate from JSX v3
 description: "Migrate from JSX v3"
-canonical: "/docs/react/latest/migrate-react"
+canonical: "/docs/react/latest/migrate-from-v3"
```
ryyppy, line 4: I am very confused by this change. Current master has a `latest/migrate-react`. So this change is essentially adding a new page `migrate-from-v3` (which I agree is a way better url, although I'd probably call it `migrate-from-jsx-3` instead).

Now that the change is out there, ppl may still refer to it, so we might need a redirection to this page via the nextjs config.

This is quite an unrelated change, so I'd want this to be a separate PR to master.

PR #654, opened 2023-03-02
Add Juspay Logo to trusted User

adhi-27:
At Juspay, we unify payment gateways to give a seamless, secure, reliable, end-to-end, enterprise-grade payment stack to our merchants. We have been using Rescript for 2 years now, with a Monorepo for a dashboard with over 5000 modules


cristianoc (approved):
Welcome!

PR #658, opened 2023-03-11
Add Instapainting.com to users section.

chrischen:
Instapainting.com is using Rescript to render its new front end stack, which is currently being used on some top search landing pages and the Ukrainian artists page.

https://www.instapainting.com/pet-portraits
https://www.instapainting.com/baby-paintings
https://www.instapainting.com/artists/ukraine


cristianoc (approved):
Welcome!

Issue #153, opened 2021-01-01
Document GADTs


ryyppy:
This one is ridiculously hard to describe in an intuitive way, so this will definitely go into an advanced section.

Not sure what practical examples would fit for e.g. UI development... IMO it's a confusing type construct and oftentimes ends up in confusing type errors, so we'd just document it for completeness reasons.

quinn-dougherty:
[Reference example](https://forum.rescript-lang.org/t/variant-primitive-type-alias/3060/3)

Josef-Thorne-A:
I wrote a tutorial on the forums for community feedback, with the intention that this could be a starting point for some official documentation:
https://forum.rescript-lang.org/t/gadt-tutorial-for-feedback/6938

fhammerschmidt:
Looks good! Feel free to submit a PR and please put it under the "Advanced Features" section.

PR #663, opened 2023-03-20
add usage `fst` and `snd` for tuple

Mng12345:


ryyppy:
Nice find, but these are part of Pervasives and I don't think we'll keep them for the future when fully migrating to rescript-core.

ryyppy:
Not that I know of, but generally the concepts of more advanced features like GADT are so abstract from a use-case and usability perspective that we didn't even bother covering them in the docs (but we actually have an open issue for them in #153, in case someone has a good idea on how to describe them in a meaningful way).

PR #667, opened 2023-04-04
Docs for the new directive decorator

zth:


cristianoc (approved):

fhammerschmidt:
Can you add it to the syntax lookup as well?

fhammerschmidt:
Woops, this IS the syntax lookup 😅 
I think what I am missing is that the https://rescript-lang.org/docs/manual/latest/attribute page does not link to the syntax lookup while only the syntax lookup page lists all of the decorators / extension points. Will do a PR myself soon.

fhammerschmidt:
here it is: https://github.com/rescript-association/rescript-lang.org/pull/668

PR #672, opened 2023-04-12
Blog post explaining new variant runtime representation

zth:


ryyppy:
Very good post! Only suggestion I'd have is to shorten the intro (two new features in v11, X and Y, this post will explain why they are important) and proceed with the sections that contain the code snippets and practical use cases.

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,383 @@
+---
+author: rescript-team
+date: "2023-04-15"
+title: Better interop with customizable variants
+badge: roadmap
+description: |
+  A tour of new capabilities coming in ReScript v11
+---
+
+ReScript v11 is around the corner, and it comes packed with new features that will improve interop with JavaScript/TypeScript. We will shine some light on each significant feature in a series of blog posts.
```
ryyppy, line 10: I would not mention any series. Just ship the posts as you see fit.

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,383 @@
...
+
+Interop with JavaScript/TypeScript is a top priority for ReScript. Recently we've made some changes to the runtime representation of variants that'll allow you to use variants for a large number of new interop scenarios, zero cost. This is important, because variants is _the_ feature of ReScript, enabling great data modeling, pattern matching and more.
```
ryyppy, line 12: This feels way too long. Just Make a bullet list of the two changes we will introduce, and explain the impact in the example sections (that's what you already do anyways, so this feels redundant).

The quicker the reader comes to meaningful examples, the better.

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,383 @@
...
+
+In addition it'll also greatly improve debuggability, because the runtime representation will map cleanly to the variant type you've defined.
+
+### Unboxed (untagged) variants
+
+We also introduce untagged variants - variants where the underlying runtime representation can be a primitive, without a specific discriminator. This will let you cleanly map to things like heterogenous array items, nullable values, and more.
+
+Now let's dive into the details.
+
+## The current state
```
ryyppy, line 40: You are already mention in the examples below how we do X in the old system. This section talks too much of the past.

The new examples highlight really well how interop issues are now solved.

Alternatively colocate parts of its content to the relevant feature sections to keep the context if needed.

zth:
@ryyppy thank you! So essentially remove the TLDR; and shorten the intro?

ryyppy (commented):
Provided some comments but obv. these are just notes on what I'd probably shorten. Not a blocker.

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,383 @@
...
+
+ReScript v11 is around the corner, and it comes packed with new features that will improve interop with JavaScript/TypeScript. We will shine some light on each significant feature in a series of blog posts.
+
+Interop with JavaScript/TypeScript is a top priority for ReScript. Recently we've made some changes to the runtime representation of variants that'll allow you to use variants for a large number of new interop scenarios, zero cost. This is important, because variants is _the_ feature of ReScript, enabling great data modeling, pattern matching and more.
+
+_Pattern matching_ is really the key here. The changes we've made will enable you to write idiomatic ReScript leveraging pattern matching in many more scenarios.
+
+These changes will ship in `v11` of ReScript, and in this post we'll tell you all you need to know about them.
+
+## TLDR;
```
ryyppy, line 18: TLDR will be unnecessary if you just get to the examples more quickly and have a simple bullet list in the intro with a few words about the highlights

zth:
@ryyppy thank you for the feedback! I've reworked the post according to your comments. Could you have a look again?

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,317 @@
...
+  Right
+
+@module("./direction.js") external myDirection: direction = "myDirection"
+```
+
+Now, this maps 100% to the TypeScript code, including letting us bring over the documentation strings so we get a nice editor experience.
+
+### String literals
+
+The same logic is easily applied to string literals from TypeScript, only here the benefits is even larger, because string literals have the same limitations in TypeScript that polymorphic variants have in ReScript.
```
cknitt, line 137:
```suggestion
The same logic is easily applied to string literals from TypeScript, only here the benefit is even larger, because string literals have the same limitations in TypeScript that polymorphic variants have in ReScript.
```

_blogposts/2023-04-15-improving-interop.mdx
```diff
@@ -0,0 +1,317 @@
...
+    }
+  | None => None
+  }
+```
+
+As you can see, you need to convert each level of nullables explicitly, which makes it hard to fully utilize pattern matching. With the new unboxed variant representation, we'll instead be able to do this:
+
+```rescript
+// The type definition below is inlined here to examplify, but this definition will live in Core and be easily accessible
+module Nullable = {
```
cknitt, line 202: The name `Nullable` may be confusing because `Js.Nullable.t` and Core's `Nullable.t` can also be `undefined`, not just `null`.

zth: Good point! Will change.

ryyppy:
@zth Thanks, way better now 👏

PR #673, opened 2023-04-13
docs: bind to a function inside an object with default export

aspeddro:
Forum post: https://forum.rescript-lang.org/t/bind-to-a-function-inside-an-object-with-export-default/4262/1


cristianoc (commented):
Screenshot 2023-04-14 at 03 55 29
6d4f0-46bf-49cc-b7b0-552e8d33ea99.png)

fhammerschmidt:
@aspeddro can you revive this one?

PR #681, opened 2023-05-14
Record type spread and coercion blog post

zth:
All feedback welcome.

### DIFF
--- a/_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
+++ b/_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
@@ -102,7 +102,7 @@ let name = nameFromB((a :> b))
 
 Notice how we _coerce_ the value `a` into the type `b`. This works because they have the same fields.
 
-Additionally, we can also coerce between records where record `a` has a subset of the fields of record `b`. The same example as above, slightly altered:
+Additionally, we can also coerce records from `a` to `b` whenever `a` is a super-set of `b` (i.e. `a` containing all the fields of `b`, and more). The same example as above, slightly altered:
 
 ```rescript
 type a = {
--- a/_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
+++ b/_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
@@ -1,0 +1,140 @@
+---
+author: rescript-team
+date: "2023-05-17"
+title: Enhancing the ergonomics of records
+badge: roadmap
+description: |
+  A tour of new capabilities coming in ReScript v11
+---
+
+[Records](https://rescript-lang.org/docs/manual/latest/record) are a fundamental part of ReScript, representing an efficient and expressive way to model structured data. Beyond offering a clear and concise definition of complex data structures, records bring numerous advantages such as strong type checking, immutability by default, great error messages, and support for exhaustive pattern matching. This allows you to write robust and maintainable code, while the compiler ensures you've handled all possible cases when dissecting a record.
+
+However, there's always room for improvement. In ReScript v10 we added support for [optional record fields](https://rescript-lang.org/docs/manual/latest/record#optional-record-fields) to make constructing records with many optional fields easier and more ergonomic. Now we're focusing on how we can improve the workflow around defining records.
+
+In ReScript v11, we're delighted to introduce two new features that will significantly boost your record manipulation capabilities: Record Type Spreads and Record Type Coercion. Let's delve into what these new features offer and how they can make working with records more ergonomic.
+
+## Type Spreads
+
+Before type spreads, creating a new record that was similar or extended from another record required spelling out every single field in the new definition. This was often tedious, error-prone and made code harder to maintain, especially when working with records with many fields.
+
+In ReScript v11, you can now spread one or more record types into a new record type. It looks like this:
+
+```rescript
+type a = {
+  id: string,
+  name: string,
+}
+
+type b = {
+  age: int
+}
+
+type c = {
+  ...a,
+  ...b,
+  active: bool
+}
+```
+
+`type c` will now be:
+
+```rescript
+type c = {
+  id: string,
+  name: string,
+  age: int,
+  active: bool,
+}
+```
+
+Keeping it as straightforward as possible, spreads are essentially a 'copy-paste' operation for fields from one or more records to another, inlining the fields from the spread records into the new record.
+
+This is going to be a much more ergonomic experience when working with types with many fields, where variations of the same underlying type are needed.
+
+### Use case: extending the built in DOM nodes
+
+This feature can be particularly useful when extending DOM nodes. For instance, in the case of the animation library Framer Motion, one could easily extend the native DOM types with additional properties specific to the library, leading to a more seamless and type-safe integration.
+
+This is how you could bind to a `div` in Framer Motion with the new record type spreads:
+
+```rescript
+type animate = {} // definition omitted for brevity
+
+type divProps = {
+  ...JsxDOM.domProps,
+  initial?: animate,
+  animate?: animate,
+  whileHover?: animate,
+  whileTap?: animate,
+}
+
+module Div = {
+  @module("framer-motion") external make: divProps => Jsx.element = "div"
+}
+```
+
+You can now use `<Div />` as a `<motion.div />` component from Framer Motion. And your type definition is now simple and easy to maintain.
+
+## Type Coercion
+
+Type coercion introduces an extra layer of flexibility when working with records. Now, records of the same shape can be coerced between each other. This means that we can cast record `a` to be record `b` at the type level, if they contain the same exact fields. An example:
+
+```rescript
+type a = {
+  name: string,
+  age: int,
+}
+
+type b = {
+  name: string,
+  age: int,
+}
+
+let nameFromB = (b: b) => b.name
+
+let a: a = {
+  name: "Name",
+  age: 35,
+}
+
+let name = nameFromB((a :> b))
+```
+
+Notice how we _coerce_ the value `a` into the type `b`. This works because they have the same fields.
+
+Additionally, we can also coerce between records where record `a` has a subset of the fields of record `b`. The same example as above, slightly altered:
+
+```rescript
+type a = {
+  id: string,
+  name: string,
+  age: int,
+  active: bool,
+}
+
+type b = {
+  name: string,
+  age: int,
+}
+
+let nameFromB = (b: b) => b.name
+
+let a: a = {
+  id: "1",
+  name: "Name",
+  age: 35,
+  active: true,
+}
+
+let name = nameFromB((a :> b))
+```
+
+Notice how `a` now has more fields than `b`, but we can still coerce `a` to `b` because `b` has a subset of the fields of `a`.
+
+### Coercing is explicit
+
+To maintain the robustness of our type system, coercing is an explicit action. Records are still nominal, preserving the same guarantees as before. You can't accidentally pass record `a` as a `b` without explicitly coercing between them. This is crucial as it prevents accidental dependencies on shapes rather than records, ensuring type safety by forcing explicit coercion between types.
+
+## Conclusion
+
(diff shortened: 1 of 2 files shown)


_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
+---
+author: rescript-team
+date: "2023-05-17"
+title: Enhancing the ergonomics of records
+badge: roadmap
+description: |
+  A tour of new capabilities coming in ReScript v11
+---
+
+[Records](https://rescript-lang.org/docs/manual/latest/record) are a fundamental part of ReScript, representing an efficient and expressive way to model structured data. Beyond offering a clear and concise definition of complex data structures, records bring numerous advantages such as strong type checking, immutability by default, great error messages, and support for exhaustive pattern matching. This allows you to write robust and maintainable code, while the compiler ensures you've handled all possible cases when dissecting a record.
```
ryyppy, line 10: A few points here:
I think most folks should be familiar with records, so no need to advertise them so heavily.

I assume the major message here is that there are records, and structurally typed objects. The latter is more convenient to use due to their ability for subtyping. Records didn't have that ergonomics so fare, but they are generally nicer to use due to its nominal typing.

To encourage more record usage, we made the way of defining records and coercing (sub-typing) records easier so we don't have to retreat to structured objects?

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+```rescript
+type c = {
+  id: string,
+  name: string,
+  age: int,
+  active: bool,
+}
+```
+
+Keeping it as straightforward as possible, spreads are essentially a 'copy-paste' operation for fields from one or more records to another, inlining the fields from the spread records into the new record.
```
ryyppy, line 50: What happens if I spread into a record that overrides an attribute?

```res
module Name = {
  type t
}

type a = {
  name: string
}

type b = {
  age: int
}

type c = {
  ...a,
  ...b,
  name: Name.t
}
```

does this error? \cc @cristianoc

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+This feature can be particularly useful when extending DOM nodes. For instance, in the case of the animation library Framer Motion, one could easily extend the native DOM types with additional properties specific to the library, leading to a more seamless and type-safe integration.
+
+This is how you could bind to a `div` in Framer Motion with the new record type spreads:
+
+```rescript
+type animate = {} // definition omitted for brevity
+
+type divProps = {
+  ...JsxDOM.domProps,
```
ryyppy, line 64: Where does the `JsxDOM` module come from? maybe worth adding a comment what this is (I am not familiar with it... I guess it has something to do with the jsx mode).

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+  age: int,
+  active: bool,
+}
+```
+
+Keeping it as straightforward as possible, spreads are essentially a 'copy-paste' operation for fields from one or more records to another, inlining the fields from the spread records into the new record.
+
+This is going to be a much more ergonomic experience when working with types with many fields, where variations of the same underlying type are needed.
+
+### Use case: extending the built in DOM nodes
```
ryyppy, line 54: Nice example!

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+module Div = {
+  @module("framer-motion") external make: divProps => Jsx.element = "div"
+}
+```
+
+You can now use `<Div />` as a `<motion.div />` component from Framer Motion. And your type definition is now simple and easy to maintain.
+
+## Type Coercion
+
+Type coercion introduces an extra layer of flexibility when working with records. Now, records of the same shape can be coerced between each other. This means that we can cast record `a` to be record `b` at the type level, if they contain the same exact fields. An example:
```
ryyppy, line 80: Maybe worth noting that type coercion has already been a thing for polymorphic variants so far. In ReScript 11, we extended _type-safe_ coercion to records as well.

(special note on type-safe, since TypeScript only does unsafe type coercion with the `as` operator)

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+  name: "Name",
+  age: 35,
+}
+
+let name = nameFromB((a :> b))
+```
+
+Notice how we _coerce_ the value `a` into the type `b`. This works because they have the same fields.
+
+Additionally, we can also coerce between records where record `a` has a subset of the fields of record `b`. The same example as above, slightly altered:
```
ryyppy, line 105: > we can also coerce between records where record `a` has a subset of the fields of record `b`

For some reason I misread this section and thought you meant where `a` is a subset of `b`.

I guess the more direct description would be...

```suggestion
Additionally, we can also coerce records from `a` to `b` whenever `a` is a super-set of `b` (i.e. `a` containing all the fields of `b`, and more). The same example as above, slightly altered:
```

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
ryyppy: Also I realized we never covered type-coercion for structural objects in the docs...

https://rescript-lang.org/try?version=v10.1.2&code=C4TwDgpgBAhlC8UDeAoKUBEA7GBbCGAXFAM7ABOAllgOYA0amMNBx1wKAviiqJFACMEyRtjytSFajShceAGwjAoAN2JxEqdGPxFMAQSzAA9lgwNtzCQGYArLO4pFygGbHjwgBQxyNYgIBKBAA+EXQAKRIAOnljGk8MAAlKDCgAajTYXwBtHQIAXQC5FDdjT08VKEJQwICgA

```
type a = {
  "name": string,
  "age": int
}

type b = {
  "name": string 
}

let v: a = {
  "name": "Anton",
  "age": 35 
}

let foo = (arg: b) => {
  Js.log("Hi" ++ arg["name"])
}

foo((v :> b))
```

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+let nameFromB = (b: b) => b.name
+
+let a: a = {
+  id: "1",
+  name: "Name",
+  age: 35,
+  active: true,
+}
+
+let name = nameFromB((a :> b))
```
ryyppy, line 129: Maybe worth pointing out that type coercion requires wrapping parenthesis. It's the first thing that popped into my eye here and there's no mention why that is.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
```diff
@@ -0,0 +1,140 @@
...
+```
+
+Notice how `a` now has more fields than `b`, but we can still coerce `a` to `b` because `b` has a subset of the fields of `a`.
+
+### Coercing is explicit
+
+To maintain the robustness of our type system, coercing is an explicit action. Records are still nominal, preserving the same guarantees as before. You can't accidentally pass record `a` as a `b` without explicitly coercing between them. This is crucial as it prevents accidental dependencies on shapes rather than records, ensuring type safety by forcing explicit coercion between types.
+
+## Conclusion
```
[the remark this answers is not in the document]
ryyppy: I guess records are now in a similar situation as variants. They are vastly superior for error messages and pattern matching ergonomics, but they couple modules and you'd need to bring the name into the module scope so you'd be able to infer types (less flexible, more rigid). Other than that, they are almost on par with their structural counter parts; even the interop part is configurable enough.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
zth: Yes, this errors right now. Cannot have overlapping properties. I omitted saying that in the post intentionally btw, just because it's a detail.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
zth: I'll change the example to not use parenthesis.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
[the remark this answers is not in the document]
zth: I'll try and come up with a clever way to incroporate structural objects since they're of course relevant, but I'd like to avoid putting to much emphasis on them. I don't personally view objects and records as "choose either or", because objects are so far behind on usefulness, in my opinion. But I'll find a way to work it in without giving it too much room.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
zth: Check again now, I rewrote parts of the intro.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
zth: I added a comment, let me know if that's enough.

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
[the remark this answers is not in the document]
ryyppy: okay looks good!

zth:
@ryyppy if you re-read it now with fresh eyes after the latest changes, how does it look?

zth:
Let's add 2 short sentences to illustrate:
- Coercing between optional and mandatory fields
- Repeated fields are disallowed

cc @cristianoc

ryyppy:
As discussed with @zth, I restructured the text a bit.

> Coercing between optional and mandatory fields

I don't fully understand how this would look like. The only coercion I got to work was this:

```rescript
type a = {
  name: string,
  age: option<int>,
}

type b = {
  age?: int,
  name: string,
}

let test = (v: b) => {
  Js.log(v.name)
}

let a: a = {name: "Bambi", age: Some(1)}

test((a :> b))
```

For me this feels quite natural, so not sure if this example yields any benefit? Or maybe I don't understand what is meant by "coercing between optional and mandatory fields".

_blogposts/2023-05-17-enhancing-the-ergonomics-of-records.mdx
ryyppy: Added a quick sentence about it just to be safe.

cristianoc:
> As discussed with @zth, I restructured the text a bit.
> 
> > Coercing between optional and mandatory fields
> 
> I don't fully understand how this would look like. The only coercion I got to work was this:
> 
> ```rescript
> type a = {
>   name: string,
>   age: option<int>,
> }
> 
> type b = {
>   age?: int,
>   name: string,
> }
> 
> let test = (v: b) => {
>   Js.log(v.name)
> }
> 
> let a: a = {name: "Bambi", age: Some(1)}
> 
> test((a :> b))
> ```
> 
> For me this feels quite natural, so not sure if this example yields any benefit? Or maybe I don't understand what is meant by "coercing between optional and mandatory fields".

That is indeed the example of converting from mandatory to optional.

cristianoc:
"Record Type Coercion is Explicit

Records are still nominally typed, so it is not possible to pass a record a as record b without an explicit type coercion. This is crucial as it prevents accidental dependencies on shapes rather than records, ensuring predictable and more robust type checking results.
"

The explicitness of coercion is a design choice, not an implementation necessity.

Issue #691, opened 2023-05-26
Playground action buttons are too far on the right


ryyppy:
<img width="2547" alt="image" src="https://github.com/rescript-association/rescript-lang.org/assets/1121889/e3ce603b-f223-4f7b-8dad-e03aab3a7ebf">

I got a little surprised how far on the right the "Format" and "Copy Share Link" buttons are.
They should probably stick to the left-hand side instead, so they are easy to see?

\cc @aspeddro

aspeddro:
See #692

fhammerschmidt:
Close?

PR #692, opened 2023-05-26
playground: align button on left

aspeddro:
#691

### DIFF
--- a/src/Playground.res
+++ b/src/Playground.res
@@ -1107,7 +1107,7 @@ module ControlPanel = {
     | _ => React.null
     }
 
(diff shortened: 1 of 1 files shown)


ryyppy:
Mh... the alignment is a little weird. So either we move it like this:
Screenshot 2023-05-26 at 15 05 36

So that the buttons start at least on the same level as the code / after the gutter.

Ideally it should be here, but not sure how this would interact with the resizable sidebar:
Screenshot 2023-05-26 at 15 05 02

aspeddro:
It is possible to integrate with resizable divider, but this format makes it easier to calculate the height of the editor panel.

PR #696, opened 2023-06-05
Add section about limitations of scoped polymorphic types

fhammerschmidt:


cristianoc (commented):
Technically it's not let bindings but function definitions.
Wonder if one can get this across without complicating the post.

cristianoc (approved):

ryyppy (approved):

PR #698, opened 2023-06-13
Add Silq logo

bdj:
Silq is a digital freight forwarder at https://www.onesilq.com. We use Rescript (and RescriptRelay) for our web application.


cristianoc:
Welcome!

PR #699, opened 2023-06-15
Add Band Protocol logo

xxibcill:
Band Protocol is a cross-chain data oracle platform that aggregates and connects real-world data and APIs to smart contracts.

we use Rescript in our block explorer [CosmoScan](https://www.cosmoscan.io/)

repo: https://github.com/bandprotocol/cosmoscan-rescript


fhammerschmidt:
Thanks. Sorry for not merging it earlier.

PR #700, opened 2023-06-24
Docs: changed to the new syntax

abdullhakim-sami:
Vscode extenstion output :

j`hello ${name}`

deprecated: The unsafe j`$(a)$(b)` interpolation is deprecated, use string template `${a}${b}` instead.

to >>> `hello ${name}`


cristianoc (approved):
Thanks!

PR #708, opened 2023-08-19
Update object mangling mention

DZakh:
Updates documentation after the fix https://github.com/rescript-lang/rescript-compiler/pull/6354

I've tried to keep the information about the behaviour but removed mentioning it from other places to keep it less confusing.


cristianoc (approved):
Thanks!

PR #714, opened 2023-09-15
Add blogpost about uncurried mode

fhammerschmidt:


_blogposts/2023-09-15-uncurried-mode.mdx
```diff
@@ -0,0 +1,127 @@
...
+```
+
+is shorter than having to write all remaining parameters again 
+
+```rescript
+let add = (a, b) => a + b
+let addFive = (b) => add(5, b)
+```
+
+Also, currying is what enabled ReScript to have optional (`~param=?`) or default (`~param=true`) parameters.
```
cknitt, line 31: Maybe remove this sentence? It might be confusing, as we can still have optional and default parameters even in uncurried mode.

_blogposts/2023-09-15-uncurried-mode.mdx
```diff
@@ -0,0 +1,127 @@
...
+is shorter than having to write all remaining parameters again 
+
+```rescript
+let add = (a, b) => a + b
+let addFive = (b) => add(5, b)
+```
+
+Also, currying is what enabled ReScript to have optional (`~param=?`) or default (`~param=true`) parameters. 
+
+This comes at a price though. Here are some examples to show the drawbacks of currying:
```
cknitt, line 33: IMHO the worst drawback/pitfall is not listed yet: In combination with `ignore` / `let _ = ...`, uncurried can lead to unexpected behavior at runtime after adding a parameter to a function, because you are accidentally ignoring the result of a partial evaluation so that the function is not called at all.

_blogposts/2023-09-15-uncurried-mode.mdx
```diff
@@ -0,0 +1,127 @@
...
+* As ReScript could not fully statically analyze when to automatically uncurry a function over multiple files, it led to unnecessary `Curry.` calls in the emitted JavaScript code. 
+* In the standard library (`Belt`), there are both curried and uncurried versions of the same function so you were required to think for yourself when to use the uncurried version and when only the curried one will work.
+
+Those are all only some small paper cuts, but all of them are intricacies that make the language harder to learn. 
+
+
+## Uncurried mode
+
+Starting with ReScript 11, your code will be compiled in uncurried mode. Yes, there is still a way to turn it off ([see below](#how-to-switch-back-to-curried-mode)), but we have decided to already default to this behavior to make it easier for newcomers.
+In uncurried mode, the introducing example yields an error:
```
cknitt, line 63:
```suggestion
In uncurried mode, the introductory example yields an error:
```

_blogposts/2023-09-15-uncurried-mode.mdx
```diff
@@ -0,0 +1,127 @@
...
+// old 
+let myFun = (~name=?, ())
+
+// new
+let myFun = (~name=?)
+```
+
+### How to switch back to curried mode
+
+While we strongly encourage all users to switch to the new uncurried mode, it is still possible to opt-out. Just add a
```
cknitt, line 101:
```suggestion
While we strongly encourage all users to switch to the new uncurried mode, it is still possible to opt out. Just add a
```

cknitt (commented):
Excellent post! Thanks a lot! 👍

_blogposts/2023-09-15-uncurried-mode.mdx
```diff
@@ -0,0 +1,127 @@
...
+        // Here it errors with "This call is missing an argument of type (~a: int)"
+        cleanup 
+      })
+    }
+  ```
+* If you wanted explicitly uncurry a function, you needed to annotate it with the uncurried dot. 
+  ```rescript
+  (. param) => ()
+  ```
+* As ReScript could not fully statically analyze when to automatically uncurry a function over multiple files, it led to unnecessary `Curry.` calls in the emitted JavaScript code.
```
tsnobip, line 54: maybe we can clearly state here than function calls in uncurried mode are now guarantied to get compiled as simple JS function calls, which is quite nice for readability of the generated JS code but also for performance (even though it was not that significant) and sometimes even correctness when applied on bindings (`Curry` sometimes badly played with  `this` and other JS features we don't care about in rescript). 

Not sure it's worth being publicized but now you can even bind classes to simple rescript records which are sometimes much nicer to use than using an opaque type and a function:
```res
myLogger->Logger.info("foo")
// vs 
myLogger.info("foo")
```

fhammerschmidt: Thanks. I am not sure about the class suggestion either. I only added your first point. 
> function calls in uncurried mode are now guarantied to get compiled as simple JS function calls

fhammerschmidt: Can you maybe give me a full example?

cristianoc: ` (f) => f()`

is an example.
Or a more realistic example using that pattern.

tsnobip (approved):
I don't see anything more to add now. Great job @fhammerschmidt!

cristianoc (approved):

_blogposts/2023-09-15-uncurried-mode.mdx
fhammerschmidt: Sorry, I should have clarified I meant for classes as records.

PR #721, opened 2023-10-05
Document 'public' feature

fhammerschmidt:


zth (approved):
Fantastic! Could you add it to the JSON schema file in the compiler repo as well, while you're at it? https://github.com/rescript-lang/rescript-compiler/blob/167c586980d1b57d5dc4ac2f82f762d7c03ebce2/docs/docson/build-schema.json

zth:
Public doesn't work in the sources config, only on the top level of bsconfig. At least that's what was reported on the forum. Can you investigate?

fhammerschmidt:
What? There is no public field on the top level. It's a field in `sourceItem`. And `sources` is either a `sourceItem` or an array of `sourceItem`s.

PR #727, opened 2023-10-17
Playground: auto-open RescriptCore when loaded

fhammerschmidt:
This is the poor man's version of the `-open RescriptCore` bsc flag, but it works!

### DIFF
--- a/src/Playground.res
+++ b/src/Playground.res
@@ -1366,6 +1366,12 @@ let parseVersion = (versionStr: string): option<(int, int, int)> => {
   }
 }
 
+// If @rescript/core is loaded, open it automatically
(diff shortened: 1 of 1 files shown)


zth:
I guess there's no way to do it the same way as one does in a regular project? A bit worried about the locs etc.

fhammerschmidt:
I will try to fight the playground in the compiler. There is a `Clflags.open_modules`, I guess it's possible to expose that in the same way as the uncurried field.

fhammerschmidt:
Ok, here is the proper way: https://github.com/rescript-lang/rescript-compiler/pull/6446/files

PR #730, opened 2023-10-20
Adapt some pages to ReScript 11 including records and variants

fhammerschmidt:


pages/docs/manual/latest/record.mdx
```diff
@@ -394,6 +394,122 @@ var me = {
...
+```rescript
+type c = {
+  id: string,
+  name: string,
+  age: int,
+  active: bool,
+}
+```
+
+Record type spreads act as a 'copy-paste' mechanism for fields from one or more records into a new record. This operation inlines the fields from the spread records directly into the new record definition, while preserving their original properties, such as whether they are optional or mandatory. It's important to note that duplicate field names are not allowed across the records being spread, even if the fields share the same type.
```
cristianoc, line 429: "share" can be confusing?
"have" should be enough

pages/docs/manual/latest/variant.mdx
```diff
@@ -193,44 +193,266 @@ let a1 = Warrior({score: 10.5})
...
+  | Present({age: Present({ageNum: Present(ageNum)})}) => Some(ageNum)
+  | _ => None
+  }
+```
+
+> Notice how `@as` allows us to say that an unboxed variant case should map to a specific underlying _primitive_. `Present` has a type variable, so it can hold any type. And since it's an unboxed type, only the payloads `'a` or `null` will be kept at runtime. That's where the magic comes from.
+
+### Decoding and encoding JSON idiomatically
+
+With unboxed variants, we have everything we need to define a JSON type:
```
cristianoc, line 353: "untagged"

cristianoc (approved):
Looks great.
Perhaps @zth would also like to take a look.

pages/docs/manual/latest/variant.mdx
```diff
@@ -193,44 +193,266 @@ let a1 = Warrior({score: 10.5})
...
+
+```rescript
+@unboxed
+type rec json =
+  | @as(false) False
+  | @as(true) True
```
zth, line 359:
```suggestion
  | Booelan(bool)
```

pages/docs/manual/latest/variant.mdx
```diff
@@ -193,44 +193,266 @@ let a1 = Warrior({score: 10.5})
...
+The same logic is easily applied to string literals from TypeScript, only here the benefit is even larger, because string literals have the same limitations in TypeScript that polymorphic variants have in ReScript.
+
+```typescript
+// direction.ts
+type direction = "UP" | "DOWN" | "LEFT" | "RIGHT";
+```
+
+There's no way to attach documentation strings to string literals in TypeScript, and you only get the actual value to interact with.
+
+## Untagged variants
```
zth, line 306: We should add a list of what can currently be unboxed (and why), and a note that it's possible that more things could be unboxed in the future.

@fhammerschmidt would you mind adding separate issues/tasks for that type of thing? I can probably add the actual list.

pages/docs/manual/latest/variant.mdx
```diff
@@ -193,44 +193,266 @@ let a1 = Warrior({score: 10.5})
...
+    ),
+  ]),
+)
+
+let usersToJson = users => Array(users->Array.map(userToJson))
+```
+
+This can be extrapolated to many more cases.
+
+## Coercion
```
zth, line 436: We should expand this section with:
- Coercing between variants (and the constraints around that)
- Why you can sometimes coerce from variant to string/int/float, and how to think about that (runtime representation must match)
- The last additions of allowing coercing strings to unboxed variants with catch-all string cases

I'll probably think of more. @fhammerschmidt as above, you mind adding a task to the milestone? Feel free to tag me and I can put something together for the actual content.

pages/docs/manual/latest/variant.mdx
zth: And, we should add a small section on the "catch all" trick, and what kind of things that enable.

zth (approved):
Just one small comment that I think should be changed now. Outside of that I think we can merge this and just continue working on whatever else was commented on.

Also, another task - we need a section on spreading variants.

fhammerschmidt:
@zth I assigned you to some new issues.

zth:
@fhammerschmidt tests seem to fail, but otherwise this should be good to go, right?

fhammerschmidt:
I have another branch open to fix the tests. Ahould be no problem to merge for now since it is not the main branch anyway.

PR #735, opened 2023-10-23
Adapt example tests to Rescript11 + uncurried + RescriptCore

fhammerschmidt:


pages/docs/manual/latest/api/belt/hash-map.mdx
```diff
@@ -31,7 +31,7 @@ let make: (~hintSize: int, ~id: id<'key, 'id>) => t<'key, 'value, 'id>
 `make(~hintSize=10, ~id)` creates a new map by taking in the comparator and `hintSize`.
 
 ```res example
-module IntHash = Belt.Id.MakeHashable({
+module IntHash = Belt.Id.MakeHashableU({
```
zth, line 34: cc @cristianoc is this needed? This isn't pre-compiled into uncurried via `MakeHashableU`?

pages/docs/manual/latest/api/js/re.mdx
```diff
@@ -68,7 +68,7 @@ Regex literals `%re("/.../")` should generally be preferred, but `fromString` is
 let firstReScriptFileExtension = (filename, content) => {
   let result = Js.Re.fromString(filename ++ "\.(res|resi)")->Js.Re.exec_(content)
   switch result {
-  | Some(r) => Js.Nullable.toOption(Js.Re.captures(r)[1])
+  | Some(r) => Js.Re.captures(r)[1]
```
zth, line 71: This isn't in `Core` right, still need to access `Js.Re`?

pages/docs/manual/latest/api/js/re.mdx
```diff
@@ -134,10 +134,16 @@ let str = "abbcdefabh"
 let break = ref(false)
 while !break.contents {
   switch Js.Re.exec_(re, str) {
-  | Some(result) => Js.Nullable.iter(Js.Re.captures(result)[0], (. match_) => {
-      let next = Belt.Int.toString(Js.Re.lastIndex(re))
-      Js.log("Found " ++ (match_ ++ (". Next match starts at " ++ next)))
-    })
+  | Some(result) =>
+    Js.Nullable.iter(
```
zth, line 138: Is there an equivalent in the global `Nullable` module?

pages/docs/manual/latest/api/js/re.mdx
```diff
@@ -134,10 +134,16 @@ let str = "abbcdefabh"
...
+      Js.Re.captures(result)[0]->Js.Nullable.fromOption,
+      (. match_) => {
+        let next = Belt.Int.toString(Js.Re.lastIndex(re))
```
zth, line 141: Can remove `Belt`?

pages/docs/manual/latest/api/js/re.mdx
```diff
@@ -134,10 +134,16 @@ let str = "abbcdefabh"
...
+        let matchStr =
+          match_->Js.Nullable.toOption->Belt.Option.getWithDefault("")
```
zth, line 143: This could be a regular match on the `Nullable.t`, right?

pages/docs/manual/latest/api/js/re.mdx
```diff
@@ -134,10 +134,16 @@ let str = "abbcdefabh"
...
+        Js.log("Found " ++ (matchStr ++ (". Next match starts at " ++ next)))
```
zth, line 144: Console.log?

zth (commented):
Awesome stuff! 😄 I left a bunch of comments, high and low. Do with it what you want.

pages/docs/manual/latest/api/belt/hash-map.mdx
cristianoc: No the compiler does not come with 2 versions of Belt.
Having to mention a `xxxU` function in documentation is pretty much a red flag, in that it instantaneously introduces tech debt.
There won't be any `xxxU` functions in future, so this is a strange point in time to document their use.

@zth do you have any thoughts about the solution here? The thing is, Belt will beed to be replaced with a library that lives outside the compiler, or in any case alongside RescriptCore. What we are witnessing here is simply that this has not happened yet.

fhammerschmidt:
So I tried to make the tests work again which is why I adapted the API examples only to the point that they compile again. Which is why I did not adapt the API examples to use RescriptCore. But this state is completely weird and I also don't know how to handle it well.

pages/docs/manual/latest/api/belt/hash-map.mdx
zth: The mid to long term idea for Belt is to be a niche lib living outside of the compiler. This is explicit documentation for Belt, and Belt will be much less needed as Core "takes over".

I'm not sure I have a point here, more an observation. I think we can break Belt out to its own package as soon as we've shipped v11. I guess that would alleviate this particular issue.

Another observation is that Belt will be much less prevalent in docs and examples now that Core exists and we're moving everything to that. So this might not be such a big problem going forward as we think.

cristianoc: Belt in the compiler can be considered an implementation detail, that can be addressed over time.
I think we can restrict attention to user visible docs for this.

So I guess we can proceed and make a note about next steps in a separate issue.

cristianoc: Yes something like this sounds like a good approach.

Though I would try to make the compiler's result, not Belt, the source of truth already now.
As otherwise that's going to pop up somewhere, be it hovering in the editor, or some error message.

cristianoc: @cknitt  https://github.com/rescript-lang/rescript-compiler/pull/6450

fhammerschmidt:
@zth Tests are passing now, let's merge this one. As I said before, I left the API docs relatively untouched besides fixing the tests. The rest of the docs is more important now.

zth:
Go for it!

PR #704, opened 2023-07-26
Playground output result (#420)

aspeddro:
Rebase onto latest master

- Don't work with JSX automatic mode 

#420

@diogomqbm

PR #742, opened 2023-11-04
Playground output result (formerly #704 and #420)

tsnobip:
I just rebased on master @aspeddro work (#704, based on #420 work by @diogomqbm).

todo:
- [ ] make it work with JSX automatic (any idea why jsx-runtime is not defined in this case?)
- [ ] explain how to display something in the render output pane and display an error when the conditions are not met
- [ ] display console messages: I'll likely just rebind console log, info, warn, error so that they append messages to a list


fhammerschmidt:
this got merged in #875 🎉

PR #748, opened 2023-11-14
Uncurried Mode docs

fhammerschmidt:


cristianoc (approved):
Looks good.
CC @zth for additional comments.

zth (approved):
Looks good to me, we can extend it when people ask questions. I guess we don't want to document `@@uncurried.swap`?

cristianoc:
> Looks good to me, we can extend it when people ask questions. I guess we don't want to document `@@uncurried.swap`?

Not really, as there are no plans to support that feature longer term. It was just for internal debugging.

ryyppy:
Bit late, but this doc feels weird, since the whole docs should revolve around uncurried to be the default and not document it as if it would be something special.

The blog post regarding curried vs uncurried should have been enough,  or at most we could have made a guide for migrating from curried to the new uncurried behavior.

The partial application syntax should IMO have been part of "Function" instead

fhammerschmidt:
You are right, I will split it up over Function and build system docs.
The `@@uncurried` decorator can probably just be added to the syntax lookup.

A migration guide (uncurried, core, rescript.json) is in the works, btw. But I am migrating a codebase while writing it, so it will take a bit.

PR #752, opened 2023-11-14
Integrate genType as 'TypeScript Interop' chapter into main docs

### LINKED ISSUE #746: Integrate genType docs into main docs as "TypeScript interop"
genType has moved into the compiler since 10.1 and flow support has been dropped

Let's move its docs into the main section as "TypeScript interop". Remove all mentions of "flowtype" and "Reason" and update the docs.

fhammerschmidt:
Closes #746


ryyppy:
A few things I'd change now that we have the chance to make structural changes to the genType docs:

0. GenType is now a builtin tool, and we should not treat this section as if it would still be a library. Maybe the Introduction should rather be something like "ReScript & TS" which would briefly mention the general concepts and philosophy, plus the setup instructions (instead of having the "Getting started" page)
1. The general sidebar structure of genType doesn't make much sense anymore (and I regretted building it this way back then). Ideally, we'd provide more concrete sidebar items that users would be looking for when interoping with TS (as we do with "JavaScript interop". The "Usage" page for instance already has a good starting point (although some items sound a little abstract), maybe it would make sense to experiment here to branch out in different documents instead.
image
2. Same feeling for supported types as in the previous point... one could provide separate sidebar items for "Bind to TS primitive values", "Bind to TS objects", "Bind variants to TS", etc with proper examples (from/to mappings) to make it copy-pastable and easy to understand.

PR #755, opened 2023-12-05
Integrate genType as 'TypeScript Interop' chapter into main docs

fhammerschmidt:
> [ryyppy](https://github.com/ryyppy) commented [2 weeks ago](https://github.com/rescript-association/rescript-lang.org/pull/752#issuecomment-1822322276) 

A few things I'd change now that we have the chance to make structural changes to the genType docs:

0. GenType is now a builtin tool, and we should not treat this section as if it would still be a library. Maybe the Introduction should rather be something like "ReScript & TS" which would briefly mention the general concepts and philosophy, plus the setup instructions (instead of having the "Getting started" page)
1. The general sidebar structure of genType doesn't make much sense anymore (and I regretted building it this way back then). Ideally, we'd provide more concrete sidebar items that users would be looking for when interoping with TS (as we do with "JavaScript interop". The "Usage" page for instance already has a good starting point (although some items sound a little abstract), maybe it would make sense to experiment here to branch out in different documents instead. 
  - image
3. Same feeling for supported types as in the previous point... one could provide separate sidebar items for "Bind to TS primitive values", "Bind to TS objects", "Bind variants to TS", etc with proper examples (from/to mappings) to make it copy-pastable and easy to understand.


pages/docs/manual/latest/bind-to-ts-other.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+# More TypeScript interop
+
+## Imported Types
+
+It's possible to import an existing TS type as an opaque type in ReScript. For example,
+
+```res
+@genType.import("./SomeFlowTypes") type weekday
+```
```
cometkim, line 15: This is a feature that should be removed in the long term (See https://github.com/rescript-lang/rescript-compiler/issues/6196)

Perhaps we'll need to mark this as deprecated now to get there in v12.

fhammerschmidt: Maybe then we don't even cover it anymore in the new docs?
cc @ryyppy

pages/docs/manual/latest/bind-to-ts-function.mdx
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# Bind to TS Function
+
+## Functions and Function Components
+
+ReScript functions are exported as JS functions of the corresponding type.
+So for example a ReScript function `foo : int => int` is exported as a JS function from numbers to numbers.
+
+If named arguments are present in the ReScript type, they are grouped and exported as JS objects. For example `foo : (~x:int, ~y:int) => int` is exported as a JS function from objects of type `{x:number, y:number}` to numbers.
```
cometkim, line 14: I'm not sure it's a regression or breaking change, but this isn't true in v11

```res
@genType
let fn1 = (~foo, ~bar) => {
  Js.log2(foo, bar)
}
```

is compiled to

```ts
export const fn1: <T1,T2>(foo:T1, bar:T2) => void = gentypeBS.fn1;
```

This is a necessary change as there will eventually be no runtime mappers, but I personally liked that feature.

It would be nice to see it added back to the compiler core, as that feature is still useful from a js interop perspective as well as gentype.

cc @cristianoc

cometkim: So I actually got **bunch** of breaking changes from my TS/ReS hybrid project.

image

This should be mentioned in the migration guide

pages/docs/manual/latest/bind-to-ts-other.mdx
cometkim: IMO, unless we prepare a separate section for deprecated features, yes.

It can be useful sometimes. It's an awkward experience for a user learning genType to discover a feature in another codebase that doesn't exist in the docs. (often happened in bucklescript era)

cometkim:
I built the homepage and took a quick look at the outline.

This is the feedback so far.

- Mentions of runtime features should be removed. It should all be considered deprecated. (specifically `genType.import`, `genType.as`, `shims`)
- The statement about runtime mappers is also no longer true.
- Projects that use genType are also TypeScript projects. For better experience, the configuration guide should cover of `tsconfig.json` too.
- The explanation of interop behavior feels like a bit of low-level information. I suggest adding guides/practice documents from a higher-level perspective. (examples: "Dealing with conventional differences", "Bring your own pipe", "Making it more convenient in JS/TS-side", etc)

Can you update any changes you are working on? I might submit another PR to your branch.

pages/docs/manual/latest/bind-to-ts-other.mdx
cristianoc: > This is a feature that should be removed in the long term (See [rescript-lang/rescript-compiler#6196](https://github.com/rescript-lang/rescript-compiler/issues/6196))
> 
> Perhaps we'll need to mark this as deprecated now to get there in v12.

It cannot removed yet until the new proposal is implemented right?

cometkim: I think so. I haven't used that feature at all, but I don't know what it's like elsewhere.

cristianoc: So I guess it can't be deprecated yet while there's no replacement.

cometkim: But isn't it the same even if the RFC is implemented in v11.x? There is currently no "replacement" for it in the RFC.

pages/docs/manual/latest/bind-to-ts-function.mdx
cometkim: Just made https://github.com/rescript-lang/rescript-compiler/issues/6530 to resolve this cc @cristianoc

fhammerschmidt:
This new version of the PR is totally simplified since lot's of runtime conversion features were removed, FCMs do not work (according to @cometkim) and more things got deprecated.
In the future, we want to have some export=all feature for genType that we can use for the docs site and can then, in theory, show the generated TS types for every single example. 

That way we would promote genType to the seamless experience it should be.

But for now I think this is the fastest path going forward. Let's merge this soon.

PR #756, opened 2023-12-05
remove docs about @obj and @deriving jsConverter/abstract

tsnobip:
these decorators are not really needed or even supported (for jsConverter) in v11.


fhammerschmidt (approved):

fhammerschmidt:
Thanks!

PR #757, opened 2023-12-05
remove unboxed doc, point it to untagged variant

tsnobip:
main use case of @unboxed is for untagged variants that now is now a real feature properly documented.


fhammerschmidt:
Maybe add a link to the https://rescript-lang.org/blog/improving-interop blogpost in the syntax lookup.
Like [@@uncurried](https://rescript-lang.org/syntax-lookup#uncurried-decorator) does it now.

Also please add the `Available since ReScript 11.0.0.` line

fhammerschmidt:
Very nice!

PR #760, opened 2023-12-10
Add new templates urls

aspeddro:
~~Remove templates section from Landing page.~~

Add new template URLs to repo https://github.com/rescript-lang/create-rescript-app

image

### DIFF
--- a/src/layouts/LandingPageLayout.res
+++ b/src/layouts/LandingPageLayout.res
@@ -581,6 +581,38 @@ module CuratedResources = {
     },
   ]
 
+  let templates = [
+    {
+      imgSrc: "/static/nextjs_starter_logo.svg",
+      title: <>
+        <div> {React.string("ReScript & ")} </div>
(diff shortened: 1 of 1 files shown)


ryyppy:
I'd still want to highlight on the landing page what's available within create-rescript-app. How does one know what ReScript supports out of the box?

aspeddro:
@fhammerschmidt `create-rescript-app` support arguments? We can add code block like `create-rescript-app --with-nextjs`

ryyppy:
I think this card should just link to the getting started page with the relevant create-rescript-app snippet... ideally later on to an actual nextjs tutorial page that uses that npx command.

I would probably not put that install link right on the card or on the front page

aspeddro:
This can be merged when `create-rescript-app` 1.6.0 is released.


cc @cknitt @fhammerschmidt

fhammerschmidt:
Will merge this now, as the release is today.

PR #763, opened 2023-12-18
Syntax Lookup: More coercion docs

fhammerschmidt:


misc_docs/syntax/operators_type_coercion.mdx
```diff
@@ -8,11 +8,20 @@ category: "operators"
 
 The `:>` operator may be used to convert a polymorphic variant to a `string` or `int`, or convert an [object](/docs/manual/latest/object) to a type with a subset of its fields.
 
+_Since ReScript `11.0.0`_ coercion also works for converting
+- from `int` to `float`
+- from record to record with the same field(s)
```
zth, line 13:
```suggestion
- from record to record with the same field(s), or when record A is a subtype of record B
```

misc_docs/syntax/operators_type_coercion.mdx
```diff
@@ -8,11 +8,20 @@ category: "operators"
...
+- from `@unboxed` variant with only strings to `string`
+- from `string` to `@unboxed` variant that have a catch-all unboxed `string` case
+- from variant to variant when applicable
+- from variant to `string`/`int`/`float` when applicable
+- for invariant type arguments such as array payloads
```
zth, line 18:
```suggestion
- for invariant type arguments such as array payloads when the runtime representation is guaranteed to be exactly the same
```

cristianoc (approved):
nice one

fhammerschmidt:
Can't fix tests for now since I would need to install the unreleased rc.8 for it.

PR #768, opened 2024-01-01
Js.log -> Console.log

cknitt:
Use `Console.log` from Core instead of `Js.log` in the docs/examples.

(Except for the `Js` docs and `Belt` docs which I left untouched.)


zth (approved):
Nice!

PR #775, opened 2024-01-07
Update Landing Page Example

aspeddro:
- Give simple example, without react
- Output is outdated

#### Other Change
- Change the color of the constructor. Currently, it is orange, but we use orange for numbers (int, float) and strings. The contrast of orange was bad on a dark background

image

### DIFF
--- a/src/layouts/LandingPageLayout.res
+++ b/src/layouts/LandingPageLayout.res
@@ -29,35 +29,41 @@ module PlaygroundHero = {
 
   let examples = [
     {
-      res: `module Button = {
-  @react.component
-  let make = (~count: int) => {
-    let times = switch count {
-    | 1 => "once"
-    | 2 => "twice"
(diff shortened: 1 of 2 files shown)


cknitt:
I am not sure about leaving out React. It's nice to show that using React is easy in ReScript.
(Idea: Maybe we could have tabs here in the future with multiple examples - vanilla JS, React, Node, React Native, whatever.)

Also, 
* Too much piping might confuse JS users?
* For ReScript 11, we are using Core by default in the playground + create-rescript-app. So we could use the more familiar `Console.log` instead of `Js.log`.

@zth @ryyppy what do you think?

ryyppy:
Main use case is still React (even if different community members disagree on this), so I really don't feel like dropping React from the examples.

Using core for the examples works for me, as long as it is denoted with an open statement.

I like the idea of having multiple examples, but requires some thought on the UI.

PR #776, opened 2024-01-07
fix playground output text selection

### LINKED ISSUE #772: "Copy code" button in Playground for generated JS?
I find myself copying the generated JS often from the playground, but it's pretty difficult because there's something in the editor that seems to make copying code hard. Try it and you'll see what I mean. Maybe we could add a button that copies the code for you or similar?

aspeddro:
Close #772

https://github.com/rescript-association/rescript-lang.org/assets/16160544/10ae2dd1-746d-4b21-b580-0b3fa5391427

### DIFF
--- a/src/Playground.res
+++ b/src/Playground.res
@@ -1601,7 +1601,6 @@ let make = (~versions: array<string>) => {
   }
 
   let onMouseMove = e => {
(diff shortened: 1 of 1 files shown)


DZakh:
Sweet, I've just started debugging the issue and noticed that you fixed it already 🤩
It was very annoying

PR #787, opened 2024-01-11
Clean up package.json and build docs

DZakh:


package.json
```diff
@@ -42,13 +42,14 @@
     "remark-parse": "^10.0.2",
     "remark-slug": "^5.1.2",
     "remark-stringify": "^7.0.3",
+    "rescript": "^11.0.0",
```
DZakh, line 45: Move to normal dependencies because it includes runtime as well

DZakh: Also, updated to the latest version, which supports `rescript -w`

cknitt (approved):

Issue #503, opened 2022-03-17
"Equality Comparison" section in the manual


ryyppy:
Alright, so the usage of `==` and `===` is not straight-forward. There's a lot of questions that need to be fully discussed, otherwise users are left with a lot of technical questions that aren't easy to answer.

This issue serves as a brain dump on the topic, so we can write up a proper manual page that discusses all the details of doing equality checks properly.

## Differences JavaScript vs ReScript

ReScript tries to mimic JS syntax / conventions. Many eslint rulesets recommend using `===` for any kind of comparisons, because it prevents unintentional type coercion which causes a lot of weird bugs. In ReScript, unintentional type coercion is not really a problem, because the user is responsible for comparing values of the same type, but it's still good to mimic the syntax, since it's easier for JS users to constantly remember the rulesets of the operators.

For complex data structures (`objects`, etc), you'd usually need a proper equality function (`Object.is`, `deepEquals`, `shallowEquals`,...) function to compare those structures, otherwise there's a lot of factors that will impact the equality comparison. 

## ReScript `==` operator

The `==` operator in ReScript is different than in JS. Depending on specific heuristics, the compiler either compiles the equality check into a `===` comparison (which is the best case scenario) or, in case the compared types are complex (i.e. variants, poly variants, records,..), introduces a runtime call `Caml_obj.caml_equal` that will do a specific equality check. 

I am not 100% what the `Caml_obj.caml_equal` check does in detail, but the heuristics for converting `==` to `===` is the following:
- comparing `int`, `float`, `string`
- comparing a value to a variant / poly variant constructor without a payload (`user == None` / `value == #hello`)
- comparing two variant / poly-variant values that do not contain any constructors with payload
[...]

PR #801, opened 2024-01-22
docs: add docs for equality and comparison

jderochervlk:
https://github.com/rescript-association/rescript-lang.org/issues/503

### DIFF
--- a/pages/docs/manual/latest/equality-comparison.mdx
+++ b/pages/docs/manual/latest/equality-comparison.mdx
@@ -1,6 +1,6 @@
 ---
 title: "Equality and Comparison"
-description: "compare"
+description: "Handling equality and comparison checks"
 canonical: "/docs/manual/latest/equality-comparison"
 ---
 
@@ -122,4 +122,4 @@ It's 2x-4x times faster than alternative deep compare functions such as [`_.isEq
 
 Instead of using `==` you could manually use a slightly faster alternative as [fast-deep-compare](https://www.npmjs.com/package/fast-deep-equal).
 
-[This repo](https://github.com/jderochervlk/rescript-perf) has benchmarks comparing results of different libraries compared to ReScrip's built in equality function.
+[This repo](https://github.com/jderochervlk/rescript-perf) has benchmarks comparing results of different libraries compared to ReScript's built in equality function.
--- a/pages/docs/manual/latest/equality-comparison.mdx
+++ b/pages/docs/manual/latest/equality-comparison.mdx
@@ -1,0 +1,125 @@
+---
+title: "Equality and Comparison"
+description: "compare"
+canonical: "/docs/manual/latest/equality-comparison"
+---
+
+# Equality and Comparison
+
+ReScript has shallow equality `===`, deep equality `==`, and comparison operators `>`, `>=`, `<`, and `<=`.
+
+## Shallow equality
+The shallow equality operator `===` compares two values and either compiles to `===` or a `bool` if the equality is known to the compiler. 
+It behaves the same as the strict equality operator `===` in JavaScript.
+
+Using `===` will never add a runtime cost.
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res
+let t1 = 1 === 1 // true
+let t2 = "foo" === "foo" // true
+let t3 = { "foo": "bar" } === { "foo": "bar"} // false
+
+let doStringsMatch = (s1: string, s2: string) => s1 === s2
+```
+```js
+var t1 = true;
+var t2 = "foo" === "foo";
+var t3 = ({ foo: "bar" }) === ({ foo: "bar" });
+
+function doStringsMatch(s1, s2) {
+  return s1 === s2;
+}
+```
+
+</CodeTab>
+
+## Deep equality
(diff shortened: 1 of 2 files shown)


pages/docs/manual/latest/equality-comparison.mdx
```diff
@@ -0,0 +1,125 @@
...
+```
+</CodeTab>
+
+## Performance of runtime equality checks
+The runtime equality check ReScript uses is quite fast and should be adequate for almost all use cases.
+It's 2x-4x times faster than alternative deep compare functions such as [`_.isEqual`](https://lodash.com/docs/4.17.15#isEqual) or [`deep-equal`](https://www.npmjs.com/package/deep-equal).
+
+Instead of using `==` you could manually use a slightly faster alternative as [fast-deep-compare](https://www.npmjs.com/package/fast-deep-equal).
+
+[This repo](https://github.com/jderochervlk/rescript-perf) has benchmarks comparing results of different libraries compared to ReScrip's built in equality function.
```
fhammerschmidt, line 125: ~~ReScrip's~~
ReScript's

jderochervlk: Fixed.

fhammerschmidt:
@ryyppy Will merge it if you think it's good to go.

cristianoc:
Small comments: when mentioning other deep comparison alternatives, it's not immediately clear that they behave the same. One would have to check as the semantics might be different in corner cases.

cristianoc:
It would be good to know if they behave the same. This is just one-off knowledge so I don't think we need unit tests.

jderochervlk:
I re-worded some of the text around performance after running some more performance benchmarks to cover different cases.
1. when we know the objects don't match
2. when the objects are the same object `equals(a, a)`
3. When the objects are both copy of the same object `equals(Object.assign({}, a), Object.assign({})`

With the additional runs it leaves fast-deep-compare as the leader in all cases. ReScript 2x over lodash for small objects with a single key, and lodash is slightly faster than rescript for large objects.

## Performance of runtime equality checks
The runtime equality check ReScript uses is quite fast and should be adequate for almost all use cases.
For small objects it can be 2x times faster than alternative deep compare functions such as Lodash's [`_.isEqual`](https://lodash.com/docs/4.17.15#isEqual).

For larger objects instead of using `==` you could manually use a faster alternative such as [fast-deep-compare](https://www.npmjs.com/package/fast-deep-equal), or write a custom comparator function.

[This repo](https://github.com/jderochervlk/rescript-perf) has benchmarks comparing results of different libraries compared to ReScript's built in equality function.

jderochervlk:
> It would be good to know if they behave the same. This is just one-off knowledge so I don't think we need unit tests.

I've been checking using differently randomly generated JSON data and I keep getting equivalent results for each library. To really hunt for edge cases I could use fast-check to make some property based tests to throw at it and see if they all behave the same.

cristianoc:
> > It would be good to know if they behave the same. This is just one-off knowledge so I don't think we need unit tests.
> 
> 
> 
> I've been checking using different randomly generated JSON data and I keep getting equivalent results for each library. To really hunt for edge cases I could use fast-check to make some property based tests to throw at it and see if they all behave the same.

@glennsl can you think of corner cases from the tests on math or other inequalities?

glennsl (commented):
Nice docs!

A couple of things not addressed here, though, is the behaviour when comparing functions, generic, abstract and private types. More generally, the idea that the output is determined by which type is inferred, and what's known about that type, rather than the underlying data type or value, might not be entirely obvious coming from a  dynamic language.

jderochervlk:
Comparing functions will throw a runtime error, but will not show a compiler error.

```rescript
let fn = () => ()
let f2 = () => ()

try {
  Console.log(fn == f2)
} catch {
| err => Console.error(err)
}
```
```
{
  RE_EXN_ID: "Invalid_argument",
  _1: "equal: functional value",
  Error: 269 |   var b_type = typeof b;
270 |   if (a_type === "function" || b_type === "function") {
271 |     throw {
272 |           RE_EXN_ID: "Invalid_argument",
273 |           _1: "equal: functional value",
274 |           Error: new Error()
                     ^
```

Would it be possible to have the compiler show an error when trying to use `==` with functions?

jderochervlk:
Another part that throws is this section, which I am not sure what it means.
```js
var tag_a = a.TAG;
var tag_b = b.TAG;
if (tag_a === 248) {
  return a[1] === b[1];
}
if (tag_a === 251) {
  throw {
    RE_EXN_ID: "Invalid_argument",
    _1: "equal: abstract value",
    Error: new Error()
  };
}
```
`a` and `b` are the 2 arguments passed to the `equal` function.

glennsl:
> Would it be possible to have the compiler show an error when trying to use == with functions?

Not in general, unless you add something like type classes to the type system. Otherwise there's no type that can accept anything except functions.

> Searching for 248 and 251 doesn't find anything in the rescript folder in node_modules.

Those are the tags for OCaml objects and abstract data types. I don't think either are officially supported by ReScript, so you can probably just ignore them.

jderochervlk:
> This PR captures all the relevant details outlined in the original issue, so from my POV we could merge this and commit further refinements as we go.
> 
> Good job on the writing and capturing performance characteristics!

Thanks! It's ready to merge.

PR #803, opened 2024-01-25
Document tagged templates

tsnobip:
Support of tagged templates was added by rescript-lang/rescript-compiler#6250. 

This PR is meant to document this new feature that will land with v11.1.


misc_docs/syntax/decorator_taggedTemplate.mdx
```diff
@@ -0,0 +1,35 @@
...
+**Since 11.1**
+
+The `@taggedTemplate` decorator is used to bind to JavaScript tag functions.
+
+### Example
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type res
```
zth, line 17: I think the name `res` is a bit confusing since it's pretty tied to ReScript. Maybe calling it `result` instead? Also, what _is_ the actual return type of that statement? Might be nice to add some sort of return type rather than an abstract type, just for clarity, even if it doesn't 100% match the actual return from Bun.

misc_docs/syntax/decorator_taggedTemplate.mdx
```diff
@@ -0,0 +1,35 @@
...
+The `@taggedTemplate` decorator is used to bind to JavaScript tag functions.
+
+### Example
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type res 
+@module("bun") @taggedTemplate 
+external sh: (array<string>, array<string>) => promise<res> = "$"
```
zth, line 19: Maybe worth adding a link to the Bun docs for this so people can go there and get some context as to what this is.

pages/docs/manual/latest/bind-to-js-function.mdx
```diff
@@ -421,3 +421,34 @@ Currently 4 directives are supported: `null_to_opt`, `undefined_to_opt`, `nullab
...
 
 `identity` will make sure that compiler will do nothing about the returned value. It is rarely used, but introduced here for debugging purpose.
+
+## Tagged template functions
+
+**Since 11.1**
+
+**Experimental** You can easily bind to JS tagged template functions.
+All you need to do is defining a binding to a function that has two arrays as arguments, 
+the first one being an array of strings and the second can be an array of anything.
```
zth, line 431: Some suggestions:
- Link to the relevant MDN article: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates
- Highlight that this form/signature is because that's how it works in JS

My reasoning is that most people will likely have heard about tagged templates, but not necessarily know anything about the internals. So I think it's important to highlight that this is all about interop, and that this is the way that things work in JS, that's why they look like they look.

pages/docs/manual/latest/bind-to-js-function.mdx
```diff
@@ -421,3 +421,34 @@ Currently 4 directives are supported: `null_to_opt`, `undefined_to_opt`, `nullab
...
+```js 
+import * as $$Bun from "bun";
+var filename = "index.res"
+var res = await $$Bun.$`ls ${filename}` 
+```
+
+</CodeTab>
+
+Notice that it gets compiled to tagged template literals in JS, which allows
+to use JS tools that only work on the literals and not by calling directly the tag function.
```
zth, line 454: Can we link to examples of these types of tag functions here? Perhaps `gql` is a good one.

pages/docs/manual/latest/interop-cheatsheet.mdx
```diff
@@ -193,6 +194,27 @@ external join: array<string> => string = "join"
 
 </CodeTab>
 
+### Tagged template functions
+
+<CodeTab labels={["ReScript", "JS Output"]}>
+
+```res example
+type res 
+@module("bun") @taggedTemplate
```
zth, line 203: Same comment as above about type naming, return type, and linking to the Bun docs for context.

zth (commented):
Great work! I added a few minor comments, do with it what you will.

pages/docs/manual/latest/bind-to-js-function.mdx
```diff
@@ -426,29 +426,34 @@ Currently 4 directives are supported: `null_to_opt`, `undefined_to_opt`, `nullab
 
 **Since 11.1**
 
-**Experimental** You can easily bind to JS tagged template functions.
-All you need to do is defining a binding to a function that has two arrays as arguments, 
+**Experimental** You can easily bind to [JS tagged template functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates).
+Tag functions in JS expect as input an array of strings and variadic parameters for the arguments of the interpolation.
+To bind to those functions in rescript, write bindings that have two arrays as arguments,
```
fhammerschmidt, line 431: To bind to those functions in ReScript, the binding signature must have two arrays as arguments,

pages/docs/manual/latest/tagged-templates.mdx
```diff
@@ -0,0 +1,124 @@
...
+
+Tagged templates provide a special form of string interpolation, enabling the creation of template literals 
+where placeholders aren't restricted to strings. Moreover, the resulting output isn't confined solely to 
+strings either. You can take a look at the [JS documentation 
+about tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates)
+to learn more about them.
+
+## Define a tag function
+
+Tag function in ReScript have the following signature:
```
fhammerschmidt, line 19: Tag functions

data/sidebar_manual_latest.json
fhammerschmidt, line 1: Maybe revert the formatting?

fhammerschmidt (changes requested):

pages/docs/manual/latest/bind-to-js-function.mdx
```diff
@@ -421,3 +421,39 @@ Currently 4 directives are supported: `null_to_opt`, `undefined_to_opt`, `nullab
...
 
 `identity` will make sure that compiler will do nothing about the returned value. It is rarely used, but introduced here for debugging purpose.
+
+## Tagged template functions
+
+**Since 11.1**
+
+**Experimental** You can easily bind to [JS tagged template functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates).
+Tag functions in JS expect as input an array of strings and variadic parameters for the arguments of the interpolation.
+To bind to those functions in rescript, the binding signature must have two arrays as arguments,
```
zth, line 431:
```suggestion
To bind to those functions in ReScript, the binding signature must have two arrays as arguments,
```

zth (approved):
Looks great! Just one tiny suggestion.

fhammerschmidt (approved):

fhammerschmidt:
You still need to change the used version in test-examples. Unfortunately, testing uses only thebmost current version for now still.

tsnobip:
thanks guys for the review, the tests pass now!

PR #806, opened 2024-01-29
Generic JSX transform

zth:
The generic JSX transform will land in `11.1`. Here's some docs, please give feedback.

I removed the "Tips & Tricks" section because I felt it didn't add much now that we have a built in generic JSX transform. Let me know if you think otherwise.

### DIFF
--- a/misc_docs/syntax/decorator_jsx_component.mdx
+++ b/misc_docs/syntax/decorator_jsx_component.mdx
@@ -1,0 +1,44 @@
+---
+id: "jsx-component-decorator"
+keywords: ["jsx", "component", "decorator"]
+name: "@jsx.component"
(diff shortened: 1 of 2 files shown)


fhammerschmidt (approved):

fhammerschmidt:
Love it!
Delivering a feature and subsequently the docs for it should be the default! 👍

fhammerschmidt:
We may want to update this one too: https://rescript-lang.org/syntax-lookup#react-component-decorator and/or add a `jsx-component-decorator`

zth:
@fhammerschmidt decorator added. Build schema adapted here: https://github.com/rescript-lang/rescript-compiler/pull/6588

PR #809, opened 2024-02-03
add docsearch metatag version

aspeddro:
Add docsearch metag to filter search by version

References:
- https://docsearch.algolia.com/docs/docsearch-v3/#filtering-your-search
- https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags


fhammerschmidt:
I needed that three years ago. Many thanks!

fhammerschmidt (approved):

PR #810, opened 2024-02-05
Document array spreads

zth:
Available in v11.1.

### DIFF
--- a/compilers/package-lock.json
+++ b/compilers/package-lock.json
@@ -14,7 +14,7 @@
         "rescript-1000": "npm:rescript@10.0.0",
         "rescript-1010": "npm:rescript@10.1.0",
         "rescript-1100": "npm:rescript@11.0.0",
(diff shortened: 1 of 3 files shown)


zth:
Right, can't merge this until it's actually released (tests fail because of the syntax change, that needs a release to work). Let's wait a bit with this then.

fhammerschmidt:
The syntax lookup probably needs to have a `...`-chapter as well. But not sure if that can even be called spread operator there, since `...` also describes a partially applied function.

zth:
Yeah, maybe it'd be more confusing to document it as a single concept like that. We can think about it for a bit.

fhammerschmidt (approved):

fhammerschmidt:
This can be merged as well?

fhammerschmidt:
Yes, update "rescript-1110" in `compilers/package.json`

PR #811, opened 2024-02-06
Import attributes

zth:
Docs for https://github.com/rescript-lang/rescript-compiler/pull/6599

### DIFF
--- a/pages/docs/manual/latest/import-from-export-to-js.mdx
+++ b/pages/docs/manual/latest/import-from-export-to-js.mdx
@@ -93,6 +93,39 @@ var studentName = Student;
 
 </CodeTab>
 
+### Use Import Attributes
+
+**Since 11.1**
+
+[Import attributes](https://github.com/tc39/proposal-import-attributes) can be used in ReScript, as long as ReScript is configured to output ES6. You do that by passing configuration to the `@module` attribute:
+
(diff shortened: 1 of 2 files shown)


pages/docs/manual/latest/import-from-export-to-js.mdx
```diff
@@ -93,6 +93,39 @@ var studentName = Student;
...
+
+This above imports the local `./myJson.json` file, adding import attributes.
+
+This is how it works:
+1. Instead of passing a string or tuple to `@module`, pass a record.
+2. This record should have a `from` key. The value of that is where you want the module to be imported from (just like the regular string to `@module` is).
+3. It should also have a `with` key, with another record where you put all the import attributes you want emitted.
+
+Notice `\"some-exotic-identifier"` - you'll need to escape any key that's not a valid ReScript record key.
+Also notice `type_`. Since `type` is a reserved keyword in ReScript, you can use `type_` instead. It will be output as `type` in the JavaScript code.
```
fhammerschmidt, line 127: > "you can use"
isn't it more like 
"you need to use `type_` instead"
since you cannot do it any other way?

zth: You can use `type` if you escape it, but since it's what you'll use 99% of the time I wanted it to have a "shortcut".

fhammerschmidt (approved):

PR #823, opened 2024-02-27
Remove some more examples that use the old uncurry code

Darkle:
Found some more examples where it used the old `(.` to uncurry things. No longer need this in Rescript v11.


fhammerschmidt:
Thank you once again!

PR #827, opened 2024-03-23
Update faq.mdx - async/await has arrived on ReScript 10.1

micklat:
it previously said "See our answer on the [Async & Promise](https://github.com/rescript-association/rescript-lang.org/blob/master/pages/docs/manual/latest/promise) page's intro."


fhammerschmidt:
That's better, thank you!

PR #838, opened 2024-04-11
ES6 -> JavaScript module with `"module": "esmodule"`

cometkim:
As follows https://github.com/rescript-lang/rescript-compiler/pull/6709

### DIFF
--- a/pages/docs/manual/latest/import-from-export-to-js.mdx
+++ b/pages/docs/manual/latest/import-from-export-to-js.mdx
@@ -8,7 +8,7 @@ canonical: "/docs/manual/latest/import-from-export-to-js"
 
 You've seen how ReScript's idiomatic [Import & Export](import-export.md) works. This section describes how we work with importing stuff from JavaScript and exporting stuff for JavaScript consumption.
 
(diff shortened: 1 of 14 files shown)


pages/docs/manual/latest/import-from-export-to-js.mdx
```diff
@@ -57,30 +57,30 @@ Here's what the `external` does:
...
+<CodeTab labels={["ReScript", "JS Output (Module)", "JS Output (CommonJS)"]}>
 
 ```res example
 @module external leftPad: (string, int) => string = "./leftPad"
 let paddedResult = leftPad("hi", 5)
 ```
 ```js
-var LeftPad = require("./leftPad");
+import * as LeftPad from "./leftPad";
 var paddedResult = LeftPad("hi", 5);
```
cometkim, line 68: > ```res
> @module external leftPad: (string, int) => string = "./leftPad"
> let paddedResult = leftPad("hi", 5)
> ```

btw, this is a problematic example that only works with transpilers such as Babel or TypeScript (a.k.a synthetic default) and does not work with standard modules.

fhammerschmidt (approved):

fhammerschmidt:
sorry, I merged your other PR and created a conflict, can you fix it then we are good to go?

cometkim:
no problem!

PR #845, opened 2024-04-16
Make a word plural

brettcannon:


fhammerschmidt:
Thanks for improving our bad english!

fhammerschmidt:
But do not hesitate to do a bigger PR with all you can find. ;-)

brettcannon:
> But do not hesitate to do a bigger PR with all you can find. ;-)

Yeah, sorry about that. 😅 Did most of these PRs from a tablet while on parental leave over multiple days, so I played it safe w/ separate PRs instead of risking messing up a single, large one.

Issue #844, opened 2024-04-16
Playground version dropdown rendering as [object object]


yangdanny97:
All the entries under the "ReScript Version" dropdown in the playground (https://rescript-lang.org/try) are being rendered as [object object] instead of a readable string.

Tested on Chrome and Firefox, clearing the cache, etc to make sure it's not just on my end.

<img width="550" alt="Screenshot 2024-04-16 at 6 36 42 PM" src="https://github.com/rescript-association/rescript-lang.org/assets/18299205/00f0863a-6582-4620-a0f0-e095187b9625">

fhammerschmidt:
uuuhhh
Thanks for reporting. It seems to be fine in local development but not when deployed.
It probably has to do with the upgrade to NextJS 14. 

@aspeddro any idea?

fhammerschmidt:
Downgraded back to 13 for now.

PR #848, opened 2024-04-17
Upgrade next v14

aspeddro:
@fhammerschmidt can you enable deploy preview, so I can check the problem #844?


ryyppy:
authorized deployment. In the future, create branches on the actual remote branch instead of using your personal fork to prevent extra authorization from our side.

PR #850, opened 2024-04-19
Add docs for %todo and -warn-error

fhammerschmidt:
last prerequisite for the blogpost

### DIFF
--- a/pages/docs/manual/latest/build-overview.mdx
+++ b/pages/docs/manual/latest/build-overview.mdx
@@ -80,4 +80,4 @@ You may want to compile your project with stricter rules for production, than wh
 rescript -warn-error +110
 ```
 
-Here, warning number 110, which is triggered when a [`%todo`](/syntax-lookup#todo) has been found, gets promoted to an error.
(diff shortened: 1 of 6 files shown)


pages/docs/manual/latest/build-overview.mdx
```diff
@@ -69,3 +69,15 @@ If you ever get into a stale build for edge-case reasons, use:
...
+
+**Since 11.1**
+
+You may want to compile your project with stricter rules for production, than when developing. With the `-warn-error` build flag, this can easily be done, for instance in a continuous integration script. E.g.:
+
+```sh
+rescript -warn-error +110
+```
+
+Here, warning number 110, which is triggered when a [`%todo`](/syntax-lookup#todo) has been found, gets promoted to an error.
```
zth, line 83: Do we have a full list of all warnings? If so we should probably link to it.

fhammerschmidt: Yes we have one and I will link it.

zth (approved):
This is great stuff, and awesome that the flag propagation was fixed @aspeddro ! WIth it this becomes a great feature.

PR #856, opened 2024-04-21
Hide top nav on reading docs on mobile

cometkim:
The top navigation disappears when you scroll down and reappears when you scroll up. Like browsers, and many content services like Medium,

This is a necessary feature for mobile readers with a small viewport.

Also made some code changes here to add memoization, because the scroll event is highly frequant.

`React.useState` preserves the reference of the set dispatcher, but doesn't for the surrounding tuple. Adding a few memo, most children, except for components that are cheap to render, are not re-rendered.

Also no deep re-render, the appearance state is propagated via old good CSS cascading rather than React state.


cometkim:
[Screencast from 2024-04-21 13-40-11.webm](https://github.com/rescript-association/rescript-lang.org/assets/9696352/4e4ea7bf-2cf0-4543-be53-c8ddb9657148)

fhammerschmidt:
This works great, nice work! 
One nit: does not seem to be implemented on the blog page.

cometkim:
Ah, the original intention was to add it to the blog, not the API docs :joy:

cometkim:
It now works on API, Docs, and Blog.

Any other place to need it?

fhammerschmidt:
Yes! https://rescript-lang.org/syntax-lookup and https://rescript-lang.org/packages

cometkim:
@fhammerschmidt Done. Now it is easily configurable by JSX composition in `App.res`

(note I've removed some unused codes in Hooks)

fhammerschmidt (approved):

PR #863, opened 2024-05-14
Install core and start migrating files

jderochervlk:
This can be merged whenever and I can start a new PR to do more work. It's probably not going to be 1 massive PR.


fhammerschmidt:
Yes, I think we could already merge this when conflicts are resolved. It's already nice to have Core available in general.

jderochervlk:
> Yes, I think we could already merge this when conflicts are resolved. It's already nice to have Core available in general.

I just merged from master, so this should be good to go.

PR #867, opened 2024-05-24
Correct comparison operator in Array.every() method

WooWan:
Fix typo in Array.every() comparison operator
Yay, my first contribution 
```rescript
let array = [1, 2, 3, 4]

//before
Console.log(array->Array.every(num => num > 4)) // true

//after
Console.log(array->Array.every(num => num <= 4)) // true
```

### DIFF
--- a/data/api/latest/core.json
+++ b/data/api/latest/core.json
@@ -9782,7 +9782,7 @@
         "kind": "value",
         "name": "every",
         "docstrings": [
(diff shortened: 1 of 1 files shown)


fhammerschmidt:
Thank you, but the file you edited here is generated from ReScript Core's interface files:
So you would need to change it here: https://github.com/rescript-association/rescript-core/blob/main/src/Core__Array.resi#L531

WooWan:
Thanks for letting me know!
I created a new pull request in the rescript-core, I close this pull request

PR #869, opened 2024-05-24
Render React in playground

tsnobip:
Now ready for review @fhammerschmidt :)


fhammerschmidt:
Do you think the react deps should be listed here as well?
image

tsnobip:
> Do you think the react deps should be listed here as well? image

I don't think so, those are ReScript dependencies.

tsnobip:
I'm actually wondering if a Console panel is that much needed, one could just open the console of the browser and have a better rendering of the logs, what do you guys think?

fhammerschmidt:
I would keep the Console tab but maybe we can have a hybrid of both Rendering and Console. In practice you most often want to have either Console or HTML, only more convoluted examples may have both.

tsnobip:
That's an interesting idea, we could do that in a follow-up PR. We could use "output" as the name of this tab.

fhammerschmidt:
Why would we want to add a new tab now only to remove it again later? Let's just add it here. Or hide it and only reveal it in the next PR.

tsnobip:
@fhammerschmidt can you take a look, in the last commit I shipped the "Output" layout, there's the react panel first then the console output. Let me know if you prefer this.

To be honest, I still think we should rather use the built-in browser dev console for the console output.

fhammerschmidt:
this got merged in #875 🎉

PR #873, opened 2024-05-30
Update @warning doc to clarify +/-

jimmytimmons:
Being new to rescript, it wasn't obvious to me how to use the warning codes to enable / disable. Updated the doc to clarify this a bit.


fhammerschmidt:
Thanks! This is useful.

fhammerschmidt:
Would you like to add that to the @@warning decorator page as well?

jimmytimmons:
> Would you like to add that to the @@warning decorator page as well?

Oop, missed that - yeah will open another PR for that page as well. thanks!

PR #877, opened 2024-06-08
Playground: Switch to babel-parser

aspeddro:
babel is battlefield tested, always keeping up with the latest features in JavaScript

- Support bigint
 - image

- Improv function printer on console
- Remove acorn-parser


fhammerschmidt:
performance is a bit slower but it is ok IMO

PR #878, opened 2024-06-08
update Core to v1.4

aspeddro:
There remain two functions using `OCamlCompact`

```
src/vendor/Json_encode.res|9 col 24| let char = c => string(OCamlCompat.String.make(1, c))
src/vendor/Json_decode.res|44 col 5| OCamlCompat.String.get(s, 0)
```


fhammerschmidt:
Thanks. Yeah let's remove that OCamlCompat methods later

PR #879, opened 2024-06-08
Add link from legacy Js modules to their current Core equivalents.

gaspard:
Having a clear warning that Core supersedes Js helps remove some confusion for newcomers. It also avoids people trying to understand code samples that use legacy features such as |>.

### DIFF
--- a/data/api/latest/js.json
+++ b/data/api/latest/js.json
@@ -3,7 +3,7 @@
     "id": "Js",
     "name": "Js",
     "docstrings": [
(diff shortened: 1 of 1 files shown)


aspeddro:
@gaspard, thanks for the contribution!!

`json` files are generated by the document generator `@recript/tools`. I like the idea of ​​adding a deprecation message.

It's better to do this in next.js side, [here](https://github.com/rescript-association/rescript-lang.org/blob/7cfe2434e2d9daef2d30dd9dec7000dc38143489/src/ApiDocs.res#L292)

fhammerschmidt:
@aspeddro would we not need to add the deprecation messages to the compiler sources themselves?

jderochervlk:
The `Js` module docs are now legacy and can no longer be found on the site.

PR #891, opened 2024-06-20
Fix grammar in let-binding.mdx

morinokami:
Change comma to period and add 's' to 'provide' in the private let bindings section of let-binding documentation.


fhammerschmidt:
Arigato!

PR #892, opened 2024-06-28
React docs updates

### LINKED ISSUE #749: Update react docs to newest api
I added a new v0.12.0 docs version to the [rescript-11](https://github.com/rescript-association/rescript-lang.org/tree/rescript-11) branch. It only needs to be updated

Below are the relevant changes from the changelog for reference

- [x] Deprecated use\*N functions in favor of changing the signature of the main hook function.
  - For example, useEffect instead of useEffectN e.g. `useEffect3(f, (a, b, c))` -> `useEffect(f, (a, b, c))`
  - The affected hooks include `useEffect`, `useLayoutEffect`, `useCallback`, `useMemo`, `useImperativeHandle`, `useInsertionEffect`
  - With this change, it is now possible to pass any value as the second argument `'deps`. In case you pass an invalid value, you will get a warning from React at runtime. You should be using one of the following values for the dependency array:
    - 0 dependencies: `[]`
    - 1 dependency: `[a]`
    - more than 1 dependency: `(a, b, ...)`
- [x] For calling `useEffect`, `useLayoutEffect` etc. _without_ a dependency array (meaning that the effect is executed on every render), there are now separate bindings `useEffectOnEveryRender`, `useLayoutEffectOnEveryRender` etc.
- [x] Added `React.lazy_`.

fhammerschmidt:
fixes #749

### DIFF
--- a/pages/docs/react/latest/beyond-jsx.mdx
+++ b/pages/docs/react/latest/beyond-jsx.mdx
@@ -22,7 +22,7 @@ A plain React component is defined as a `('props) => React.element` function. Yo
 
 Here are some examples on how to define your own component types (often useful when interoping with existing JS code, or passing around components):
 
-```res
+```res example
 // Plain function type
 type friend = {name: string, online: bool}
 type friendComp = friend => React.element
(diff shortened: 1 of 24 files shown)


pages/docs/react/latest/beyond-jsx.mdx
```diff
@@ -38,9 +38,24 @@ The types above are pretty low level (basically the JS representation of a React
 
 A ReScript React component needs to be a (sub-)module with a `make` function and `props` type to be usable in JSX. To make things easier, we provide a `@react.component` decorator to create those functions for you:
 
-<CodeTab labels={["Decorated", "Expanded"]}>
+<div className="hidden">
```
aspeddro, line 41: Why hidden?

pages/docs/react/latest/components-and-props.mdx
```diff
@@ -116,9 +116,20 @@ function Greeting(props) {
 
 In JSX, you can apply optional props with some special syntax:
 
+<div className="hidden">
```
aspeddro, line 119: Should remove?

aspeddro (commented):
Looks great!! I left a comment

pages/docs/react/latest/beyond-jsx.mdx
fhammerschmidt: to be able to compile the code in the tests and still don't show the prelude to the user

pages/docs/react/latest/components-and-props.mdx
fhammerschmidt: Moved the prelude up a bit and mentioned that the following examples will utilize this one so I could get rid of the hidden class.

aspeddro (approved):

PR #898, opened 2024-07-25
Update OurUsers.res

arcasraak:


fhammerschmidt:
Thanks, but it appears you split up the logo and this commit into two PRs. Can you do one with both please?

PR #906, opened 2024-08-27
Gentype: Bring back shims documentation

fhammerschmidt:
This is based on the old shims docs but with code examples directly put into the docs.

### DIFF
--- a/pages/docs/manual/latest/typescript-integration.mdx
+++ b/pages/docs/manual/latest/typescript-integration.mdx
@@ -214,6 +214,55 @@ These features are for experimentation only. They could be changed/removed any t
 
 - Export object and record types as interfaces. To activate, add `"exportInterfaces": true` to the configuration. The types are also renamed from `name` to `Iname`.
 
+
+## Shims
+
+A shim is a TS file that provides user-provided definitions for library types.
+
+Required only if one needs to export certain basic ReScript data types to JS when one cannot modify the sources to add annotations (e.g. exporting ReScript lists), and if the types are not first-classed in genType.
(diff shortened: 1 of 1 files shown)


pages/docs/manual/latest/typescript-integration.mdx
```diff
@@ -214,6 +214,55 @@ These features are for experimentation only. They could be changed/removed any t
...
+      "Js": "Js",
+      "ReactEvent": "ReactEvent",
+      "RescriptPervasives": "RescriptPervasives",
+      "ReasonReact": "ReactShim"
+    },
+  },
+}
+```
+
+and add relevant `.shim.ts` files in a directory which is visible by ReScript e.g.
```
zth, line 240: This has always confused me. Can I put the shims anywhere and gentype will just automatically find the `<moduleName>.shim.ts` and import it relatively? Since we're not explicitly saying what directory to look in.

fhammerschmidt: Maybe it could, but this PR is just to restore the docs as it works currently?
We can create a separate issue for your concerns in the compiler repo, but, I think, shims can be removed entirely after we use d.ts. files, or do they not? @cometkim

cometkim: > I think, shims can be removed entirely after we use d.ts. files, or do they not? @cometkim

Yes that's the plan for v12, but I don't think it would be backported into v11

fhammerschmidt:
@cometkim do you think it's ok to merge?

cometkim:
Yep! LGTM

PR #929, opened 2024-11-06
allow static export on master

cometkim:


cometkim:
let me do merge it for infra PoC

PR #936, opened 2024-11-07
Add Zed extension

nerdalytics:
Add link to the Zed extension under community supported section


fhammerschmidt:
Oh nice. How good does it work already for you compared to VSCode or (neo)vim? I would only put it there if it's in a relatively stable state.

fhammerschmidt (approved):

PR #937, opened 2024-11-12
Automatically sync playground URL

tsnobip:
Right now, Copy Share Link button is a bit broken, it reloads the whole page.

So I used shallow replace to avoid that. Since syncing the URL was then almost instant, I thought that syncing the URL while typing or on version/module system switch would be nice too so I implemented it.


fhammerschmidt:
any reason why you deleted the Next.resi?

tsnobip:
> any reason why you deleted the Next.resi?

I guess Next.res (and before Next.re) originally had some runtime code, now it's pure bindings so the interface file is redundant and actually the interface file was even wrong because it used `let` signature instead of `external` so it generated extra code for nothing.

PR #952, opened 2024-12-06
V12 docs

fhammerschmidt:
There is now an .env file with a pointer to the current **latest** and **next** versions. Renamed the "latest" folder to "v11.0.0". When v12 is released we can just move the latest pointer to it.


fhammerschmidt:
Works fine now with the redirect! Href tests pass as well. Thanks a lot for your suggestions @cometkim 
I'd appreciate a review.

package.json
```diff
@@ -67,6 +67,7 @@
     "@mdx-js/react": "^2.3.0",
     "autoprefixer": "^10.4.14",
     "cssnano": "^6.0.1",
+    "dotenv": "^16.4.7",
```
aspeddro, line 70: Is this dependency necessary?

fhammerschmidt: Probably not if I just put the version aliases in a JS or JSON file from which all of nextjs config, code and tests can read. Shall I?

aspeddro: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#loading-environment-variables

fhammerschmidt: I know but I use the .env in a test script as well.

aspeddro (approved):

zth (approved):

PR #957, opened 2024-12-30
Fix docsearch version metatag

aspeddro:
The search doesn't work because the scraper can't
find the url with `latest` anymore.

This PR changes the meta tag, after merging I will update the crawler


fhammerschmidt:
Ah. Many thanks for fixing it.

fhammerschmidt (approved):

PR #960, opened 2025-01-03
feat: document `ref` for self-referencing

czabaj:
I encountered a problem with valid JS code that was hard to translate into rescript - the problem is [discussed here in the forum](https://forum.rescript-lang.org/t/recursion-in-a-function-argument-via-callback/5744/6) - it turned out the `ref` can help with self-referencing.

I promised to add this trick to the documentation, so here it is. Feel free to criticize 😸 

 I currently added it just to the v11 docs, once this passes validation, I will copy that into other versions as well. I assume it is useful for all versions where mutable let assignment exists?

Here is how it renders

Screenshot 2025-01-03 at 12 38 39


fhammerschmidt:
Sorry I missed this during my christmas break.

The example seems pretty heavy and we usually try to avoid React-specific code in the main documentation. Can you come up with a simpler example that doesn't need React?

jderochervlk:
@codex can you come up with a different example?

PR #965, opened 2025-01-28
Fix typo in the "Object Method" section

kantuni:
"to a JS objects" → "to JS objects"

P.S. Sorry for a messy commit history, I was using GitHub.


fhammerschmidt (approved):
No worries, those will be squashed anyway. 
And thanks again!

PR #987, opened 2025-03-13
Fix old blog post with broken images

jderochervlk:
Fixes https://github.com/rescript-lang/rescript-lang.org/issues/219

- adds missing images to static folder
- adds the blog posts folder to tailwind so it can pick up if we use classes there
- added some basic classes to make the images look a bit better, but I didn't spend too much time on an archived post
- removed author avatars that went to broken URLs and configured rendering to exclude empty ones


fhammerschmidt:
Digging up old corpses now? Well nice that it got finally fixed.

fhammerschmidt (approved):

jderochervlk:
> Digging up old corpses now? Well nice that it got finally fixed.

Hey, if it's low hanging fruit I wanna close out some issues!

Issue #981, opened 2025-03-06
Generate v12 Stdlib/Core docs


fhammerschmidt:

nojaf:
Would this also produce a `llm.txt` for core?
Was there any progress on this during the retreat?

fhammerschmidt:
This is thankfully implemented by @tsnobip!

PR #990, opened 2025-03-16
Update codebase to rescript v12.0.0-alpha.9

aspeddro:
First step to working on #981


tsnobip (approved):
great work, thanks @aspeddro !

PR #995, opened 2025-04-01
Update primitive-types.mdx

m-rinaldi:


fhammerschmidt:
Thanks!
While this suggestion works, the `String.get` from rescript-core does not return a Char either. ReScript Core has no `Char` datatype so one would need to use `OCamlCompat.String` instead.

Currently it looks like it won't be in V12 either (at least from the start) since we are almost done stabilizing the standard library and probably won't add another module until the corresponding RFC is implemented: https://github.com/rescript-lang/rescript/issues/7028

Would you be up to adapt the text so that it suggests to use OCamlCompat instead?

fhammerschmidt (approved):
Thanks!

PR #997, opened 2025-04-04
Blog post about unified operators

cometkim:


_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+date: "2025-04-05"
+title: "Rethinking Operators: How ReScript Simplifies Arithmetic and Beyond"
+badge: roadmap
+description: |
+    Discover how unified operators in ReScript v12 simplify arithmetic, reduce syntax noise — plus, a glimpse into the future roadmap.
+---
+
+## Introduction
+
+For upcoming ReScript v12, we're upgrading common arithmetic operators to be "Unified Operators"
```
fhammerschmidt, line 12: In the upcoming ReScript v12, we're upgrading common arithmetic operators to 'Unified Operators".

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+
+This post covers both the reasoning behind the change and what’s next on the roadmap. If you're interested in the implementation details, you’ll find them in [the pull request](https://github.com/rescript-lang/rescript/pull/7057).
+
+## Problems in operators
+
+Until v12, the operator syntax had a few notable problems.
+
+### Unwanted syntax gap
+
+Having different operators for each type is not familiar to JavaScript users, and not even allowing overloading can feel strange to most programmers.
```
fhammerschmidt, line 32: Using different operators for each type is unfamiliar to JavaScript users, and the lack of operator overloading can feel strange to most programmers.

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+It was inconvenient because we had to shadow the definition every time.
+
+```res
+let add_bigint = {
+  open BigInt!
+  1n + 2n
+}
+```
+
+Every time we introduce a new primitive type (who knows!), we face the same issue across all arithmetic operators.
```
fhammerschmidt, line 61: Every time we introduce a new primitive type (who knows?), we run into the same issue with all arithmetic operators.

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+Then why do we not use the same pretty operators everywhere as JavaScript?
+
+```res
+let compare_int = 1 < 2
+let compare_float = 1.0 < 2.0
+```
+
+```js
+let compare_int = true
+let compare_float = true
```
fhammerschmidt, line 74: Not sure if that is a good example since the `<` and `>` get optimized away.
However in a function with parameters it also uses `Primitive_object.lessthan`instead.

But I think you don't want to show the compiled output here, but demonstrate how you write it by hand in JS.
Also why not `camelCase`?

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+//   
+//   You can convert float to int with Belt.Float.toInt.
+//   If this is a literal, try a number without a trailing dot (e.g. 20).
+```
+
+Because ReScript only intentionally supports monomorphic operations, `(int, int) => int` in this case. Users have to perform type conversions explicitly where necessary.
+
+While it's tempting to allow full operator overloading or polymorphism like JavaScript or TypeScript, we intentionally avoid it to preserve predictable type inference and runtime performance guarantees.
+
+However, comparisons are actually exceptional ones. Let's summon the polymorphism.
```
fhammerschmidt, line 96: However, comparisons are actually the exception. Time to summon polymorphism!

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+let fn2 = (a: float, b) => a + b // (float, float) => float 
+let fn3 = (a, b: float) => a + b // (float, float) => float
+
+let inv1 = (a: int, b: float) => a + b  // => (int, int) => int
+//                                   ^ error: cannot apply float here, expected int
+```
+
+Then, in IR, it is translated to the corresponding compile-time primitive based on the unified type.
+
+This approach has been inspired by an awesome language [F#](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/arithmetic-operators#operators-and-type-inference), which is also originated from OCaml.
```
fhammerschmidt, line 142: This approach is inspired by the awesome language [F#](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/arithmetic-operators#operators-and-type-inference), which also originates from OCaml.

_blogposts/2025-04-05-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+
+### Reduced internal complexity
+
+By normalizing how primitive operators are added and managed, it also lowers maintenance overhead. A couple of new operators are actually being added by new community contributors [@MiryangJung](https://github.com/MiryangJung) and [@gwansikk](https://github.com/gwansikk)
+
+### Support most JavaScript operators
+
+We are working to support more unified operators to close the syntax gap with JavaScript.
+
+In the ReScript v12, most familiar JavaScript operators should work as is. Not only arithmetic operators, but also bitwise and shift operators.
```
fhammerschmidt, line 162: In ReScript v12, most familiar JavaScript operators should work as-is — not just arithmetic operators, but also bitwise and shift operators.

fhammerschmidt (changes requested):

cometkim:
I've revised the section about the roadmap on comparison operators to avoid confusion

fhammerschmidt:
Example function names should all be camel case.

cometkim:
Done, and updated the date

_blogposts/2025-04-11-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+date: "2025-04-11"
+title: "Rethinking Operators: How ReScript Simplifies Arithmetic and Beyond"
+badge: roadmap
+description: |
+  Discover how unified operators in ReScript v12 simplify arithmetic, reduce syntax noise — plus, a glimpse into the future roadmap.
+---
+
+## Introduction
+
+In the upcoming ReScript v12, we're upgrading common arithmetic operators to "'Unified Operators".
```
fhammerschmidt, line 12: `"'Unified Operators"` has a superfluous `'`

_blogposts/2025-04-11-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+  open BigInt!
+  1n + 2n
+}
+```
+
+Every time we introduce a new primitive type (who knows?), we run into the same issue with all arithmetic operators.
+
+### Hidden risk of polymorphism
+
+Then why do we not use the same pretty operators everywhere as JavaScript?
```
fhammerschmidt, line 65: So why don’t we just use the same pretty operators everywhere, like in JavaScript?

_blogposts/2025-04-11-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+### Support most JavaScript operators
+
+We are working to support more unified operators to close the syntax gap with JavaScript.
+
+In ReScript v12, most familiar JavaScript operators should work as-is — not just arithmetic operators, but also bitwise and shift operators.
+
+- Remainder operator (`%`) - [#7152](https://github.com/rescript-lang/rescript/pull/7152)
+- Exponentiation operator (`**`) - [#7153](https://github.com/rescript-lang/rescript/pull/7153)
+- Bitwise operators (`~`, `^`, `|`, `&`) - [#7172](https://github.com/rescript-lang/rescript/issues/7172)
+- Shift operators (`<<`, `>>`, `>>>`) - [#7171](https://github.com/rescript-lang/rescript/issues/7171)
```
fhammerschmidt, line 173: Are we sure we can land them or should we remove this one?

fhammerschmidt (changes requested):

_blogposts/2025-04-11-introducing-unified-operators.mdx
cometkim: I can handle it

cometkim:
Need more updates?

_blogposts/2025-04-11-introducing-unified-operators.mdx
```diff
@@ -0,0 +1,268 @@
...
+---
+
+## Introduction
+
+In the upcoming ReScript v12, we're upgrading common arithmetic operators to "Unified Operators".
+
+This means that we can now use a single infix operator syntax for multiple numeric types, and even for string concatenation.
+
+```res
+let add_int = 1 + 2
```
fhammerschmidt, line 17: remaining snake_cases

fhammerschmidt (approved):

PR #999, opened 2025-04-08
Update interop-cheatsheet.mdx

shuklatushar12219829:
had a small mistake: the return type of filter's callback should be bool, not 'b.


fhammerschmidt (approved):
Thanks!

PR #1002, opened 2025-04-23
update documentation about lazy values

tsnobip:
Update documentation about lazy values after we dropped the lazy keyword and updated the Lazy API (rescript-lang/rescript#7399)


fhammerschmidt:
Thanks a lot!

fhammerschmidt (approved):

PR #1003, opened 2025-05-01
Hide focus outline in documentation search bar

mediremi:
The documentation search input already had an `outline-none` Tailwind class, but due to conflicting styles in the same stylesheet with higher specificity, it had no effect.

By changing it to `focus:outline-none`, this rule now has high enough specificity to actually disable outlines.

I've also increased the input's height to 3.5rem, which makes it much easier to click on.

| Before | After |
| -- | -- |
| image | image

### DIFF
--- a/styles/_docsearch.css
+++ b/styles/_docsearch.css
@@ -89,7 +89,7 @@
 }
 
 .DocSearch-Input {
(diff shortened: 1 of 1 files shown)


fhammerschmidt (approved):
Very nice, thank you!

PR #1009, opened 2025-05-03
Document spreads in the syntax lookup

zth:
I think I got them all.


misc_docs/syntax/language_spreads.mdx
```diff
@@ -0,0 +1,95 @@
+---
+id: "spreads"
+keywords: ["spread", "record", "variant", "partial"]
```
zth, line 3: Not sure about the keywords.

fhammerschmidt: looks fine to me

fhammerschmidt:
array, list and polymorphic variants have spreads as well

PR #1014, opened 2025-05-03
Editor page + reanalyze guide

fhammerschmidt:
@zth you can add docs about @editorCompletefrom etc. on the new "Editor" page which is actually the extended editor-plugins page.


jderochervlk (approved):
Looks good!

PR #1016, opened 2025-05-04
Add pages about the ReScript Association and donations

fhammerschmidt:


fhammerschmidt:
I guess we also need the imprint page with links to the statutes?
https://github.com/rescript-association/website/blob/7bf16291beccd95c7b0cb44aa1cf4567f08c8d07/pages/imprint.mdx

ryyppy:
yes we need the imprint (in the footer?)

fhammerschmidt:
Yes, the Footer can grow a bit I guess.

PR #1028, opened 2025-05-05
remove deadcode related to nav dropdown

jderochervlk:
Cleans up some code that is no longer needed and solves a bug that causes the mobile nav to disappear when scrolling on the doc page.


fhammerschmidt (approved):

cometkim:
Can you please let me know what problem what we had because of collapsable navbar?

The thick persistent header is really annoying to me because I often read on mobile :sweat:

PR #1030, opened 2025-05-05
update installation section

jderochervlk:
- Updates recommended node version to 20 to avoid suggesting versions that are end of life
- Added Deno to package managers
- Added that create-rescript-app will setup Next.js or Vite with React and Tailwind CSS to make it clear what it does
- Removed Core from manual setup steps
- Changed "Babel and Webpack" to "Vite, Rspack, Rollup" to use modern tools as examples


fhammerschmidt:
As discussed, either remove deno for now or we can also hold this one off until create-rescript-app supports it.

jderochervlk:
@fhammerschmidt I removed the instructions for Deno from create-rescript-app for now. It's doable, but I'll need to add Deno there first.

fhammerschmidt (approved):

PR #1031, opened 2025-05-06
Adding a sub nav to the docs section to reduce the amount of content in the main nav

jderochervlk:
This will also let us move things out of the doc sidebar and create subnav links for: Getting Started, Guides, Api, Syntax Lookup, and anything else we want to split up to reduce the amount of things in the sidebar. This should also help increase visibility into different sections and provide a better path for newcomers.

An example of what I am attempting:
image

## Desktop
image

## Tablet
image

## Mobile
image

## Wide desktop
image


fhammerschmidt (approved):
Looks great now!

PR #1034, opened 2025-05-08
Bring back collapsible navbar on mobile layout

cometkim:
Has been removed from https://github.com/rescript-lang/rescript-lang.org/pull/1028, but I guess it was not intentional..? Or let me know if there was any problem, I will fix


fhammerschmidt:
Superseded by #1035

PR #1038, opened 2025-05-18
Use same-origin resources for compiler bundles.

cometkim:
Compiler JS bundles are uploaded into a Cloudflare R2 bucket.

https://github.com/rescript-lang/rescript/pull/7472

Serving resources from the same-origin removes CORS preflight when visiting the playground and prevents charges for R2 reads.

I made the `sync-bundles` script to download all resources from the bucket.

And we have two options:

1. Commit resources to git
   - Only new versions can be additionally downloaded.
   - The repo will get significant bloat.
2. Don't commit resources to git
   - For developers, they need to sync it at least once.
   - The sync must be run per deployment, which will significantly slow down.

Option 2 is not feasible for Vercel deployments, but for Cloudflare Pages.

Alternatively, we can run a [custom deployment with GitHub Actions](https://developers.cloudflare.com/workers/ci-cd/external-cicd/github-actions/) (When we use Cloudflare Workers instead of Pages). Option 2 + `@actions/cache` would be the best.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 *.res linguist-language=ReScript
 *.resi linguist-language=ReScript
+
+/public/playground-bundles/**  binary linguist-vendored
--- a/.gitignore
+++ b/.gitignore
@@ -26,9 +26,12 @@ lib/
 .vercel
 
 src/**/*.mjs
-scripts/**/*.mjs
+scripts/gendocs.mjs
+scripts/generate_*.mjs
 
 # Generated via generate-llms script
 public/llms/manual/**/llm*.txt
 public/llms/react/**/llm*.txt
 pages/docs/**/**/llms.mdx
+
+public/playground-bundles/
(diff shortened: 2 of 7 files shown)


fhammerschmidt (approved):
Then I'd say check them in until Vercel is gone and do the proper action setup later?

cometkim:
Once checked in, it cannot be undone. Therefore, to avoid bloating the git repository, we shouldn't check in forever.

cometkim:
This will work on Cloudflare pages anyway after changing the command to `npm run sync-bundles && npm run build`.

cometkim:
Then we switch to Cloudflare Workers. I'll add GitHub Actions to optimize it and get the versioned snapshot deployments and fork previews we want.

cometkim:
Ok, I'm merging it; this won't break the existing Vercel build.

cometkim:
Whoa, option 2 isn't as bad as I expected. Full sync takes only ~20s in the Cloudflare Pages build.

PR #1044, opened 2025-05-24
Vlk v11 split

jderochervlk:
This won't actually be merged in, but it will just become a `v11` branch once it's ready and serve as the source for `https://v11.rescript-lang.org`


cometkim:
Ahh I totally forgot. I will work on it this weekend

jderochervlk:
> Ahh I totally forgot. I will work on it this weekend

No worries! I'll try and wrap this up with links and such. I think the home page and playground shouldn't be on the subdomain and versioned, so I'll delete more and set up some redirects. We should probably also add a no-index for seo and remove the LLM stuff from older versions.

PR #1046, opened 2025-06-04
Document the new exception API

### LINKED ISSUE #1032: Exceptions documentation
There were some recent changes on master which renamed some things for Exceptions.

Can someone who actually understands this please update https://rescript-lang.org/docs/manual/v12.0.0/exception accordingly?

Thanks

tsnobip:
Fixes #1032.

### DIFF
--- a/pages/docs/manual/v12.0.0/async-await.mdx
+++ b/pages/docs/manual/v12.0.0/async-await.mdx
@@ -130,15 +130,15 @@ You may use `try / catch` or `switch` to handle exceptions during async executio
 ```res
 // For simulation purposes
 let authenticate = async () => {
-  raise(Exn.raiseRangeError("Authentication failed."))
+  JsError.RangeError.throwWithMessage("Authentication failed.")
 }
 
 let checkAuth = async () => {
   try {
     await authenticate()
   } catch {
-  | Exn.Error(e) =>
-    switch Exn.message(e) {
+  | JsExn(e) =>
+    switch JsExn.message(e) {
     | Some(msg) => Console.log("JS error thrown: " ++ msg)
     | None => Console.log("Some other exception has been thrown")
     }
@@ -152,14 +152,14 @@ You may unify error and value handling in a single switch as well:
 
 ```res
 let authenticate = async () => {
-  raise(Exn.raiseRangeError("Authentication failed."))
+  JsError.RangeError.throwWithMessage("Authentication failed.")
 }
 
 let checkAuth = async () => {
   switch await authenticate() {
   | _ => Console.log("ok")
-  | exception Exn.Error(e) => 
-    switch Exn.message(e) {
+  | exception JsExn(e) => 
+    switch JsExn.message(e) {
     | Some(msg) => Console.log("JS error thrown: " ++ msg)
     | None => Console.log("Some other exception has been thrown")
     }
--- a/pages/docs/manual/v12.0.0/exception.mdx
+++ b/pages/docs/manual/v12.0.0/exception.mdx
@@ -420,9 +420,9 @@ try {
 }
 ```
 
-### Throw a value that is not an error
+### Throw a value that is not an JS Error
 
-For some reason, if you need to throw something that is not a JS error, use `JsExn.throw`:
+If you want to throw any value that is not a valid JS Error, use `JsExn.throw`:
 
 <CodeTab labels={["ReScript", "JS Output"]}>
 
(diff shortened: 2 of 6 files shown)


pages/docs/manual/v12.0.0/async-await.mdx
```diff
@@ -130,15 +130,15 @@ You may use `try / catch` or `switch` to handle exceptions during async executio
...
+  JsError.RangeError.throwWithMessage("Authentication failed.")
 }
 
 let checkAuth = async () => {
   try {
     await authenticate()
   } catch {
-  | Exn.Error(e) =>
-    switch Exn.message(e) {
+  | JsExn(e) =>
```
nojaf, line 140: It is still a little confusing where `JsExn` is coming from.
Is this the "Exceptions are just a special kind of variant" from the beginning?

pages/docs/manual/v12.0.0/async-await.mdx
```diff
@@ -152,14 +152,14 @@ You may unify error and value handling in a single switch as well:
...
-  raise(Exn.raiseRangeError("Authentication failed."))
+  JsError.RangeError.throwWithMessage("Authentication failed.")
 }
 
 let checkAuth = async () => {
   switch await authenticate() {
   | _ => Console.log("ok")
-  | exception Exn.Error(e) => 
-    switch Exn.message(e) {
+  | exception JsExn(e) =>
```
nojaf, line 161: Is this exhaustive? How does this work with the other built-in exceptions?
Do we cover all of them with this?
Do I add `exception Invalid_argument` above this case when I want to match that?

pages/docs/manual/v12.0.0/exception.mdx
```diff
@@ -434,6 +420,36 @@ try {
 }
 ```
 
+### Throw a value that is not an error
+
+For some reason, if you need to throw something that is not a JS error, use `JsExn.throw`:
```
fhammerschmidt: If you want to throw any value that is not a valid JS Error, use `JsExn.throw`:

fhammerschmidt (changes requested):

pages/docs/manual/v12.0.0/async-await.mdx
```diff
@@ -130,15 +130,15 @@ You may use `try / catch` or `switch` to handle exceptions during async executio
 ```res
 // For simulation purposes
 let authenticate = async () => {
-  raise(Exn.raiseRangeError("Authentication failed."))
+  JsError.RangeError.throwWithMessage("Authentication failed.")
```
nojaf, line 133: Are `JsError` and `JsExn` the same thing?

pages/docs/manual/v12.0.0/async-await.mdx
tsnobip: I just updated this page so it stays in sync with the new API, I think this could be improved in another PR.

pages/docs/manual/v12.0.0/async-await.mdx
tsnobip: This is already covered in the exceptions page, I fear we would repeat ourselves too much otherwise.

pages/docs/manual/v12.0.0/async-await.mdx
tsnobip: `JsError` is the module around JS errors while `JsExn` is the module around things that can be thrown and caught in JS, so pretty much anything actually.

tsnobip:
@nojaf 

> Thank you very much for tackling this. I can't deny that it is still a very confusing area of the language.

Well, I really tried my best but I think the previous situation was even worse, but of course we're open to any idea that could further improve the situation!

> Could you also link to the Promise API docs in `async-await.mdx`? 

I didn't specifically update the `async-await.mdx` page, I just wanted to make sure we stopped referencing the deprecated API, I think improvements of this page should come in a different PR.

> The example of what to do with `Promise.catch` on https://rescript-langorg-git-update-exceptions-rescript-association.vercel.app/docs/manual/v12.0.0/api/core/promise#value-catch is out of sync I think.

The Stdlib/Core API hasn't been synced for ages, as soon as this (#981) is done, it will be corrected since the docstring is fixed already.

pages/docs/manual/v12.0.0/async-await.mdx
nojaf: That is one of looking at it, but there you assume your user has read that exceptions page.
And so you assume the docs are read in a certain order.
This is not my experience with the docs. I often go and look at documentation because I have a specific problem, so I find links between these pages very valuable.

tsnobip: @nojaf you're totally right, I'd just change this a subsequent PR, I just replaced the calls to the `Exn` module.

nojaf: Thanks Paul!

PR #1053, opened 2025-06-17
Update and rename pull-request.yml to ci.yml

fhammerschmidt:
Trigger a build if github-actions[bot] pushes a commit on master (new api docs)


fhammerschmidt:
ok that is the wrong spot
deployment needs to be handled at cloudflare

PR #1060, opened 2025-06-26
Fix rescript dependency / upgrade to v12 alpha.14

cknitt:


tsnobip:
looks good but I already merged it in [afc345f9e2d9cef0bae795ddd26af7bcc2b669bb](https://github.com/rescript-lang/rescript-lang.org/commit/afc345f9e2d9cef0bae795ddd26af7bcc2b669bb)

PR #1069, opened 2025-07-07
Fix compilation error pattern matching

mediremi:
This fixes the parsing and pattern matching of compilation errors in the playground, which was changed as part of https://github.com/rescript-lang/rescript-lang.org/pull/1066


aspeddro (approved):
Thank you!!

PR #1075, opened 2025-07-21
Correctly sort experimental compiler versions available in playground

mediremi:
The existing sort logic assumed that there are only up to 10 dev/alpha/beta/rc releases per version, but v12 had 15 alphas.

| Before | After |
| ------ | ----- |
| image | image |


fhammerschmidt (approved):
Thanks!

PR #1078, opened 2025-07-28
Upgrade tailwind v4

jderochervlk:
I just followed the migration guide.
https://tailwindcss.com/docs/upgrade-guide


fhammerschmidt:
the header now disappears when scrolling down even on desktop

jderochervlk:
> the header now disappears when scrolling down even on desktop

Fixed!

PR #1081, opened 2025-08-12
Add prettier and format everything

jderochervlk:
The only default I didn't use was semicolons. Not using them makes the JS look more like ReScript, which I think is nice.


fhammerschmidt:
Stick to the full defaults, I'd say. 
Especially for the JS examples in the markdown files.

It makes ReScript stand out more as well ;-)

fhammerschmidt:
Also what do you think about adding a format check in CI? Can be in another PR of course.

jderochervlk:
> Also what do you think about adding a format check in CI? Can be in another PR of course.

I added a CI check and a pre-commit hook. I noticed we didn't have anything set up for hooks, so I added Lefthook.

https://lefthook.dev/

fhammerschmidt:
I think that's overkill. Can we rather nudge devs to have the correct formatter executed on save?

jderochervlk:
> I think that's overkill. Can we rather nudge devs to have the correct formatter executed on save?

Sure, I can remove the pre-commit hook.

PR #1082, opened 2025-08-16
Enable auto-run by default in playground UI

haydenflinner:
One less click to get to the output of a playground link (click Output vs Checking AutoRun then clicking Output) and the accompanying very-cool live-edit experience.

I spent a few hours trying to get "Output as the default tab, with Results filled in once page loads and can run the code" working seamlessly too (i.e. 0 clicks), but am struggling with waiting for the output tab's iframe to be loaded before triggering the "RunCode" functionality.

### DIFF
--- a/src/common/CompilerManagerHook.res
+++ b/src/common/CompilerManagerHook.res
@@ -446,7 +446,7 @@ let useCompilerManager = (
                 errors: [],
                 result: FinalResult.Nothing,
                 logs: [],
-                autoRun: false,
+                autoRun: true,
                 validReactCode: false,
               }))
             | Error(errs) =>


fhammerschmidt:
@tsnobip what do you think?

I am just worried about a full refresh giving you the same result when you accidentally produce an infinite loop. The page will freeze then.

tsnobip:
@fhammerschmidt I'm not a big fan, it means if the code is too heavy and blocks the page, you just can't modify it, it doesn't seem like a great default to me. We should at least add a query param to disable it, or run the code in the background? But can you actually do it if you use react? Or maybe we should add some kind of timer and force stop the execution after some time. I don't know, what do you guys think?

fhammerschmidt:
React needs to run on the main (UI) thread, so a web worker would only help for everything else.
So I'd go for the extra query parameter. 

@haydenflinner Would it be enough for you if you could bookmark `https://rescript-lang.org/try?autorun` or something similar?
And if so, are you keen to implement it?

haydenflinner:
That would work but I won't be implementing right now. Thanks for reviewing guys!

PR #1093, opened 2025-08-27
Draft: let? blog post

fhammerschmidt:


cristianoc:
Assistant:

>Coming from other languages:

>Rust developers will immediately recognize and appreciate it
>JavaScript/TypeScript developers might find it unfamiliar but useful
>Functional programming advocates may prefer more general monadic solutions

>Overall Prediction
>I think this will be well-received once people try it. The ReScript community generally values pragmatism alongside functional programming principles, and this strikes a good balance. The experimental flag is smart - it gives the team room to iterate based on real usage patterns.
The main debates will likely center around:

>Whether to expand beyond built-ins to custom variants
Syntax bikeshedding (though let? is quite good)
Integration with future language features

>This feels like a natural evolution for ReScript - maintaining its OCaml roots while improving developer experience for common patterns.

looks spot on to me

cristianoc:
>Areas for Improvement
>Structure and flow: The post jumps between concepts without clear transitions. The section about "two observations" feels disconnected from what follows, and the explanation of how it works could be better integrated.
Terminology inconsistency: The post uses "let-unwrap," "let?," and "unwrap syntax" interchangeably without establishing clear naming conventions upfront.
Technical precision: Some statements could be clearer:

>"matches the explicitness of ReScript we all like" - what does this mean exactly?
"plays nicely with async/await" - how specifically?
The distinction between block/local vs top-level usage needs clarification
"And it is for block or local bindings only; top-level usage is rejected" - why this restriction?

>Suggested Improvements

>Lead with a clear definition of what let? does before showing examples
Structure as problem → solution → implementation rather than jumping between topics
Standardize terminology early and stick with it
Add clearer section breaks to help readers follow the flow
Expand on the restrictions - explain why certain limitations exist
Include error cases in examples to show the full picture

Seems legit, except the last suggestion might be too much

cristianoc:
Btw `let?` is great balanced design. Nice one @zth

_blogposts/2025-09-01-let-unwrap.mdx
```diff
@@ -0,0 +1,205 @@
...
+    }
+  }
+```
+
+Two observations:
+
+1. with every `switch` expression, this function gets nested deeper.
+2. The `Error` branch of every `switch` is just an identity mapper (neither wrapper nor contents change).
+
+The only alternative in ReScript was always to use some specialized methods:
```
cknitt:
```suggestion
The only alternative in ReScript was always to use some specialized functions:
```

_blogposts/2025-09-01-let-unwrap.mdx
```diff
@@ -0,0 +1,205 @@
...
+```res
+let getUserPromises = id =>
+  fetchUser(id)
+  ->Result.flatMapOkAsync(user => Promise.resolve(user->decodeUser))
+  ->Result.flatMapOkAsync(decodedUser => ensureUserActive(decodedUser))
+```
+
+**Note**: `Result.flatMapOkAsync` among some other async result helper functions are brand new in ReScript 12 as well!
+
+This is arguably better, more concise, but also harder to understand because we have two wrapper types here, `promise` and `result`. And we have to wrap the non-async type in a `Promise.resolve` in order to stay on the same type level. Also we are giving up on our precious `async`/`await` syntax here.
```
cknitt: Also, this has implications on performance (two additional calls to helper functions).

It might be interesting to compare the JS output for the version with helper functions vs. the let-unwrap function to illustrate this.

_blogposts/2025-09-01-let-unwrap.mdx
```diff
@@ -0,0 +1,205 @@
...
+}
+```
+
+This strikes a balance between conciseness and simplicity that we really think fits ReScript well.
+
+With `let?`, we can now safely focus on the the happy-path in the scope of the function. There is no nesting as the `Error` is automatically mapped. But be assured the error case is also handled as the type checker will complain when you don't handle the `Error` returned by the `getUser` function.
+
+This desugars to a **sequence** of early-returns that you’d otherwise write by hand, so there’s **no extra runtime cost** and it plays nicely with `async/await` as the example above suggests.
+
+Of course, it also works for `option` with `Some(...)`.
```
cknitt, line 153: Might be worth mentioning that you cannot mix `option` with `result` though.

_blogposts/2025-09-01-let-unwrap.mdx
fhammerschmidt: Added more JS output.

_blogposts/2025-09-01-let-unwrap.mdx
```diff
@@ -0,0 +1,291 @@
...
+The only alternative in ReScript was always to use some specialized functions:
+
+```res
+let getUser = id =>
+  fetchUser(id)
+  ->Result.flatMapOkAsync(user => Promise.resolve(user->decodeUser))
+  ->Result.flatMapOkAsync(decodedUser => ensureUserActive(decodedUser))
+```
+
+**Note**: `Result.flatMapOkAsync` among some other async result helper functions are brand new in ReScript 12 as well!
```
nojaf, line 56: Maybe add a link to https://rescript-lang.org/docs/manual/v12.0.0/api/stdlib/result#value-flatMapOkAsync

nojaf (approved):
This looks good and I think it’s ready to go.

One feeling I still get is that it reads like a new idea we haven't really tested in our codebase. I'm not sure if that's an issue or worth mentioning, but it's hard to tell whether this will be game-changing or not.

_blogposts/2025-09-01-let-unwrap.mdx
```diff
@@ -0,0 +1,291 @@
...
+</CodeTab>
+
+## Experimental features
+
+We have added an `experimental-features` infrastructure to the toolchain. If you use the new build system that comes with ReScript 12 by default, you can enable it in `rescript.json` like so:
+
+```json
+{
+  "experimental-features": {
+    "letUnwrap": true
```
benadamstyles, line 274:
```suggestion
    "LetUnwrap": true
```

jderochervlk:
We should probably get this into a place to publish.

PR #1096, opened 2025-09-03
Add GADT documentation to the manual

Josef-Thorne-A:
This adds an initial version of documentation for GADTs with a thorough tutorial.
I edited this from my original forum post for clarity and added a practical
example -- writing bindings to Node's stream API.﻿

Feedback and further refinement welcome!
fixes: #153

### DIFF
--- a/pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
+++ b/pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
@@ -91,14 +91,13 @@ we try to return a standard timezone from this function. Actually, this seems li
 we still want to be able to match on all cases of the variant sometimes, and a naive attempt at this will not pass the type checker. A naive example will fail:
 
 ```res example
-let convertToDaylight = tz => {
+let convertToDaylight = tz =>
   switch tz {
   | EST => EDT
   | CST => CDT
   | CDT => CDT
   | EDT => EDT
   }
-}
 ```
 
 This will complain that `daylight` and `standard` are incompatible. To fix this, we need to explicitly annotate to tell the compiler to accept both:
@@ -107,11 +106,14 @@ This will complain that `daylight` and `standard` are incompatible. To fix this,
 let convertToDaylight : type a. timezone<a> => timezone<daylight> =  // ...
 ```
 
-`type a.` here defines a _locally abstract type_ which basically tells the compiler that the type parameter a is some specific type, but we don't care what it is. The cost of the extra specificity and safety that GADTs give us is that the compiler is not able to help us with type inference as much.
+The syntax `type a.` here defines a _locally abstract type_ which basically tells the compiler that the type parameter a is some specific type, but we don't care what it is. The cost of the extra specificity and safety that
+GADTs give us is that the compiler less able to help us with type inference.
 
 ## Varying return type
 
-Sometimes, a function should have a different return type based on what you give it, and GADTs are how we can do this in a type-safe way. We can implement a generic `add` function that works on both `int` or `float`:
+Sometimes, a function should have a different return type based on what you give it, and GADTs are how we can do this in a type-safe way. We can implement a generic `add` function[^1] that works on both `int` or `float`:
+
+[^1]: In ReScript v12, the built-in operators are already generic, but we use them in this example for simplicity.
 
 ```res example
 type rec number<_> = Int(int): number<int> | Float(float): number<float>
@@ -131,7 +133,9 @@ let bar = add(Int(1), Float(2.0)) // the compiler will complain here
 
 How does this work? The key thing is the function signature for add. The `number` GADT is acting as a _type witness_. We have told the compiler that the type parameter for `number` will be the same as the type we return -- both are set to `a`. So if we provide a `number<int>`, `a` equals `int`, and the function will therefore return an `int`.
 
-We can also use this to avoid returning `option` unnecessarily. This example is adapted from Real World Ocaml, chapter 9. We create an array searching function can be configured to either raise an exception, return an `option`, or provide a `default` value depending on the behavior we want.
+We can also use this to avoid returning `option` unnecessarily. We create an array searching function which either raises an exception, returns an `option`, or provides a `default` value depending on the behavior we ask for.[^2]
+
+[^2]: This example is adapted from [here](https://dev.realworldocaml.org/gadts.html).
 
 ```res example
 module If_not_found = {
@@ -171,7 +175,7 @@ In an advanced case that combines the above techniques, we can use GADTs to sele
 The below example defines a `num` type similar to our above addition example, but this lets us use `int` and `float` arrays
 interchangeably, hiding the implementation type rather than exposing it. This is similar to a regular variant. However, it is a tuple including embedding a `numTy` and another value.
 `numTy` serves as a type-witness, making it
-possible to recover type information that was hidden dynamically. Matching on `numTy` will "reveal" the type of the other value in the pair.We can use this to write a generic sum function over arrays of numbers:
+possible to recover type information that was hidden dynamically. Matching on `numTy` will "reveal" the type of the other value in the pair. We can use this to write a generic sum function over arrays of numbers:
 
 ```res example
 type rec numTy<'a> =
@@ -197,8 +201,8 @@ Javascript libraries that are highly polymorphic or use inheritance can benefit 
 of Node's `Stream` API.
 
 This API has a method for binding event handlers, `on`. This takes an event and a callback. The callback accepts different parameters
-depending in which event we are binding to. A naive implementation might look similar to this, defining a
-separate method for each stream event to wrap the unsafe version of on.
+depending on which event we are binding to. A naive implementation might look similar to this, defining a
+separate method for each stream event to wrap the unsafe version of `on`.
 
 ```res example
 module Stream = {
@@ -223,7 +227,7 @@ the type signature of the callback and pass this instead of a plain string.
 Additionally, we use some type parameters to represent the different types of Streams.
 
 This example is complex, but it enforces tons of useful rules. The wrong event can never be used
-with the wrong callback, but it also will never be used with the wrong kind of stream. The compiler will complain for example if we try to use a `Pipe` event with anything other than a `writable` stream.
+with the wrong callback, but it also will never be used with the wrong kind of stream. The compiler will for example complain if we try to use a `Pipe` event with anything other than a `writable` stream.
 
 The real magic happens in the signature of `on`. Read it carefully, and then look at the examples and try to
 follow how the type variables are getting filled in, write it out on paper what each type variable is equal
@@ -242,7 +246,7 @@ module Stream = {
   @unboxed
   type chunk =
     | Str(string)
-    // Node uses actually its own buffer type, but for the tutorial are just using the stdlib's buffer type.
+    // Node uses actually its own buffer type, but for the tutorial we are using the stdlib's buffer type.
     | Buf(buffer)
 
   type rec event<_, _> =
@@ -287,7 +291,7 @@ encounter will be more difficult to understand, and the compiler sometimes requi
 type your code.
 
 However, there are definite situations where GADTs are the _right_ decision
-and will _simplify_ your code and help you avoid bugs, even rendering some bugs impossible. The `Stream` example above is a good example where the "simpler" alternative of using regular variants or even strings.
+and will _simplify_ your code and help you avoid bugs, even rendering some bugs impossible. The `Stream` example above is a good example where the "simpler" alternative of using regular variants or even strings
 would lead to a much more complex and error prone interface.
 
 Ordinary variants are not necessarily _simple_ therefore, and neither are GADTs necessarily _complex_.


fhammerschmidt:
Sorry for the format check shenanigans - should be fixed when you update to the latest master commit now.
Will do a thorough review later tonight.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
+---
+title: "Generalized Algebraic Data Types"
+description: "Generalized Algebraic Data Types in Rescript"
+canonical: "/docs/manual/v12.0.0/generalized-algebraic-data-types"
+---
+
+# Generalized Algebraic Data Types
+
+Generalized Algebraic Data Types (GADTs) are an advanced feature of Rescript's type system. "Generalized" can be somewhat of a misnomer -- what they actually allow you to do is add some extra type-specificity to your variants. Using a GADT, you can give the individual cases of a variant _different_ types.
```
fhammerschmidt: `ReScript's`

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+For a quick overview of the use cases, reach for GADTs when:
+
+1. You need to distinguish between different members of a variant at the type-level
```
fhammerschmidt: Either add a `.` here or remove them all in the enumeration.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+2. You want to "hide" type information in a type-safe way, without resorting to casts.
+3. You need a function to return a different type depending on its input.
+
+GADTs usually are overkill, but when you need them, you need them! Understanding them from first principles is difficult, so it is best to explain through some motivating examples.
```
fhammerschmidt, line 17: Hell yeah this really helps a lot.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
fhammerschmidt: also remove the `-` from `type-level`

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+  | EST => EDT
+  | CST => CDT
+  | EDT | CDT /* or, _ */ => failwith("Invalid timezone provided!")
+  }
+}
+```
+
+This function is only valid for a subset of our variant type's constructors but we can't handle this in a type-safe way using regular variants. We have to enforce that at runtime -- and moreover the compiler can't help us ensure we are failing only in the invalid cases. We are back to dynamically checking validity like we would in a language without static typing. If you work with a large variant type long enough, you will frequently find yourself writing repetitive catchall `switch` statements like the above, and for little actual benefit. The compiler should be able to help us here.
+
+Lets see if we can find a way for the compiler to help us out with normal variants. We could define another variant type to distinguish the two kinds of timezone.
```
fhammerschmidt: Let's see if we can find a way for the compiler to help us with normal variants.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+```
+
+This has a lot of problems. For one, it's cumbersome and redundant. We would now have to pattern-match twice whenever we deal with a timezone that's wrapped up here. The compiler will force us to check whether we are dealing with daylight or standard time, but notice that there's nothing stopping us from providing invalid timezones to these constructors:
+
+```res example
+let invalid_tz1 = Daylight(EST)
+let invalid_tz2 = Standard(EDT)
+```
+
+Consequently, we still have to write our redundant catchall cases. We could define daylight savings time and standard time as two _separate_ types, and unify those in our `daylight_or_standard` variant. That could be a passable solution, but that makes a distinction really would like to do is implement some kind of _subtyping_ relationship. We have two _kinds_ of timezone. This is where GADTs are handy:
```
fhammerschmidt: change
> That could be a passable solution, but that makes a distinction really would like to do is implement some kind of subtyping relationship.

to

> That could be a passable solution, but what we would really like to do is implement some kind of subtyping relationship.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+```res example
+let convert_to_daylight = tz => {
+  switch tz {
+  | EST => EDT
+  | CST => CDT
+  | CDT => CDT
+  | EDT => EDT
+  }
+}
```
fhammerschmidt: Nit: Save a line by removing the outer brackets

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+```
+
+This will complain that `daylight` and `standard` are incompatible. To fix this, we need to explicitly annotate to tell the compiler to accept both:
+
+```res example
+let convert_to_daylight : type a. timezone<a> => timezone<daylight> =  // ...
+```
+
+`type a.` here defines a _locally abstract type_ which basically tells the compiler that the type parameter a is some specific type, but we don't care what it is. The cost of the extra specificity and safety that GADTs give us is that the compiler is not able to help us with type inference as much.
```
fhammerschmidt: the type parameter `a`  (add backticks)

fhammerschmidt: The cost of the extra specificity and safety that GADTs give us is that the compiler is less able to help us with type inference.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+## Varying return type
+
+Sometimes, a function should have a different return type based on what you give it, and GADTs are how we can do this in a type-safe way. We can implement a generic `add` function that works on both `int` or `float`:
```
fhammerschmidt: Just a sidenote: Since ReScript 12 this is possible with the relaxed arithmetic operators which now work on int and float (and string) together. So it might not be the best example but I realize it is for demonstrative purposes.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+let foo = add(Int(1), Int(2))
+
+let bar = add(Int(1), Float(2.0)) // the compiler will complain here
+```
+
+How does this work? The key thing is the function signature for add. The number GADT is acting as a `type witness`. We have told the compiler that the type parameter for `number` will be the same as the type we return -- both are set to `a`. So if we provide a `number<int>`, `a` equals `int`, and the function will therefore return an `int`.
+
+We can also use this to avoid returning `option` unnecessarily. This example is adapted from Real World Ocaml, chapter 9. We create an array searching function can be configured to either raise an exception, return an `option`, or provide a `default` value depending on the behavior we want.
```
fhammerschmidt: We have a rule to never mention OCaml on this site. 
https://github.com/rescript-lang/rescript-lang.org/blob/master/CONTRIBUTING.md#technical-writing-documentation

the exception would be a copyright situation of course and I saw that RWO states [they want to be attributed](https://github.com/realworldocaml/book/blob/master/LICENSE.md) for their examples.

So ideally either adapt this example to be different enough or mention it in a footnote below the example instead of the prose text.

fhammerschmidt: We create an array searching function that can be configured to either ...

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+```
+
+## Hide and recover Type information Dynamically
+
+In a very advanced case that combines many of the above techniques, we can use GADTs to selectively hide and recover type information. This helps us create more generic types.
+The below example defines a `num` type similar to our above addition example, but this lets us use `int` and `float` arrays
+interchangeably, hiding the implementation type rather than exposing it. This is similar to a regular variant. However, it is a tuple including embedding a `num_ty` and another value.
+`num_ty` serves as a type-witness, making it
+possible to recover type information that was hidden dynamically. Matching on `num_ty` will "reveal" the type of the other value in the pair.We can use this to write a generic sum function over arrays of numbers:
```
fhammerschmidt: missing space before `We`

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+  switch witness {
+  | Int => Num(Int, array->Array.reduce(0, add_int))
+  | Float => Num(Float, array->Array.reduce(0., add_float))
+  }
+}
+```
+
+## A Practical Example -- writing bindings:
+
+Javascript libraries that are highly polymorphic or use inheritance can benefit hugely from GADTs, but they can be useful for bindings even in other cases. The following examples are writing bindings to a simplified
```
fhammerschmidt, line 200: The following examples show how to write bindings to a simplified version of Node's `Stream` API.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+of Node's `Stream` API.
+
+This API has a method for binding event handlers, `on`. This takes an event and a callback. The callback accepts different parameters
+depending in which event we are binding to. A naive implementation might look similar to this, defining a
```
fhammerschmidt: depending on which event ...

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+separate method for each stream event to wrap the unsafe version of on.
```
fhammerschmidt: wrap on in backticks

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+Stream like `Readable` or `Writable`, which have all sorts of special interactions with the callback that jeopardize our type-safety, we are going to be in even deeper trouble.
+
+Instead, we can use the same GADT technique that let us vary return type to vary the input type.
+Not only are we able to now just use a single method, but the compiler will guarantee we are always using the correct callback type for the given event. We simply define an event GADT which specifies
+the type signature of the callback and pass this instead of a plain string.
+
+Additionally, we use some type parameters to represent the different types of Streams.
+
+This example is complex, but it enforces tons of useful rules. The wrong event can never be used
+with the wrong callback, but it also will never be used with the wrong kind of stream. The compiler will will complain for example if we try to use a `Pipe` event with anything other than a `writable` stream.
```
fhammerschmidt: The compiler will for example complain ...

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+
+  type writable
+  type readable
+
+  type buffer = {buffer: ArrayBuffer.t}
+
+  @unboxed
+  type chunk =
+    | Str(string)
+    // Node uses actually its own buffer type, but for the tutorial are just using the stdlib's buffer type.
```
fhammerschmidt: ... but for the tutorial we are ...

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+  @send
+  external on: (t<'inputStream>, event<'inputStream, 'callback>, 'callback) => unit = "on"
+
+}
+
+let writer = Stream.Writable.make()
+let reader = Stream.Readable.make()
+// Types will be correctly inferred for each callback, based on the event parameter provided
+writer->Stream.on(Pipe, r => {
+  Js.log("Piping has started")
```
fhammerschmidt: please use `Console.log` instead

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+writer->Stream.on(End, _ => Js.log("End reached"))
+
+```
+
+This example is only over a tiny, imaginary subset of node's Stream API, but it shows a real-life example
+where GADTs are all but indispensable.
+
+## Conclusion
+
+While GADTs can make your types extra-expressive and get more safety, with great power comes great
```
fhammerschmidt: While GADTs can make your types extra-expressive and provide more safety, with great power comes great responsibility.

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+responsibility. Code that uses GADTs can sometimes be too clever for its own good. The type errors you
+encounter will be more difficult to understand, and the compiler sometimes requires extra help to properly
+type your code.
+
+However, There are definite situations where GADTs are the _right_ decision
```
fhammerschmidt: However, there

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+and will _simplify_ your code and help you avoid bugs, even rendering some bugs impossible. The `Stream` example above is a good example where the "simpler" alternative of using regular variants or even strings.
```
fhammerschmidt: remove `.`

pages/docs/manual/v12.0.0/generalized-algebraic-data-types.mdx
```diff
@@ -0,0 +1,294 @@
...
+would lead to a much more complex and error prone interface.
+
+Ordinary variants are not necessarily _simple_ therefore, and neither are GADTs necessarily _complex_.
+The choice is rather which tool is the right one for the job. When your logic is complex, the highly expressive nature of GADTs can make it simpler to capture that logic.
+When your logic is simple, it's best to reach for a simpler tool and avoid the cognitive overhead.
+The only way to get good at identifying which the situation calls for is to try out
```
fhammerschmidt: The only way to get good at identifying which tool to use in a given situation is to practice and experiment with both.

Josef-Thorne-A:
Thanks for the feedback -- I think I managed to get everything. Regarding the RWO reference, I removed any direct mention of OCaml -- just a quick footnote linking the RWO chapter.

fhammerschmidt (approved):
Great stuff! Thank you so much!

PR #1099, opened 2025-09-03
make playground bundle endpoint customizable

cometkim:
combo with https://github.com/rescript-lang/rescript/pull/7834


fhammerschmidt:
Need this already so I am merging.

PR #1103, opened 2025-09-05
Show how to use GADTs with inline function type annotations

mediremi:
I think using inline function type annotations makes the examples clearer.


fhammerschmidt (approved):
Yes that reads clearer, thanks!

YawarRaza7349:
You might also want to add some text (a sentence or two) introducing this alternative notation then, since it's currently introduced with the separated-out notation, and the reader might wonder why it switched without explanation.

PR #1116, opened 2025-09-28
refactor: add order and section to language manual frontmatter

jderochervlk:
This is in preparation for the migration to React Router.

This doesn't do anything yet, but frontmatter will be the primary way to add details and organize markdown files instead of the build scripts and config files we use today.


fhammerschmidt:
What if I don't use the order at all for a section? Will it be sorted alphabetically?

jderochervlk:
> What if I don't use the order at all for a section? Will it be sorted alphabetically?

That could be done pretty easily. I can tweak that, but this is a good starting point that reduces some of the file changes in the main PR.

PR #1117, opened 2025-10-03
Document the deprecation of @raises over @throws

tsnobip:
Following the changes introduced by rescript-lang/rescript#7932

### DIFF
--- a/misc_docs/syntax/decorator_does_not_throw.mdx
+++ b/misc_docs/syntax/decorator_does_not_throw.mdx
@@ -1,0 +1,15 @@
+---
+id: "does-not-throw-decorator"
+keywords: ["doesNotThrow", "decorator"]
+name: "@doesNotThrow"
+summary: "This is the `@doesNotThrow` decorator."
+category: "decorators"
+---
+
+> This decorator requires [`reanalyze`](https://github.com/rescript-lang/reanalyze), a code analysis tool for ReScript, to be installed. [Click here to read about how you get started with reanalyze.](https://github.com/rescript-lang/reanalyze).
+
+`@doesNotThrow` is used to override the reanalyze's exception analysis and state that an expression does not throw any exceptions, even though the analysis reports otherwise. This can happen for example in the case of array access where the analysis does not perform range checks but takes a conservative stance that any access could potentially throw.
+
+### References
+
+- [Reanalyze: Exception Analysis](https://github.com/rescript-lang/reanalyze/blob/master/EXCEPTION.md)
--- a/misc_docs/syntax/decorator_does_not_throw.mdx
+++ b/misc_docs/syntax/decorator_does_not_throw.mdx
@@ -6,10 +6,10 @@ summary: "This is the `@doesNotThrow` decorator."
 category: "decorators"
 ---
 
-> This decorator requires [`reanalyze`](https://github.com/rescript-lang/reanalyze), a code analysis tool for ReScript, to be installed. [Click here to read about how you get started with reanalyze.](https://github.com/rescript-lang/reanalyze).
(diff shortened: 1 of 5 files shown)


misc_docs/syntax/decorator_does_not_throw.mdx
```diff
@@ -0,0 +1,15 @@
+---
+id: "does-not-throw-decorator"
+keywords: ["doesNotThrow", "decorator"]
+name: "@doesNotThrow"
+summary: "This is the `@doesNotThrow` decorator."
+category: "decorators"
+---
+
+> This decorator requires [`reanalyze`](https://github.com/rescript-lang/reanalyze), a code analysis tool for ReScript, to be installed. [Click here to read about how you get started with reanalyze.](https://github.com/rescript-lang/reanalyze).
```
fhammerschmidt: can you link to https://rescript-lang.org/docs/manual/v12.0.0/editor-plugins#code-analysis instead?

fhammerschmidt: the section might need an update as well though, but it probably can be just copied over from the reanalyze Readme. (and then use `throws` instead of `raises`).

fhammerschmidt (changes requested):

pages/docs/manual/v12.0.0/editor-plugins.mdx
```diff
@@ -40,7 +40,89 @@ The code analysis provides extra checks for your ReScript project, such as detec
...
+
+- The libraries currently modeled are limited to `Array`, `Buffer`, `Bytes`, `Char`, `Filename`, `Hashtbl`, `List`, `Pervasives`, `Str`, `String` from the standard library, and `bs-json`, and `Json` from `Js`. Models are currently vendored in the analysis, and are easy to add (see [`src/ExnLib.ml`](src/ExnLib.ml))
+- Generic exceptions are not understood by the analysis. For example `exn` is not recognized below (only concrete exceptions are):
+
+```rescript
+try (foo()) { | exn => throw(exn) }
+```
+
+- Uses of e.g. `List.hd` are interpreted as belonging to the standard library. If you re-define `List` in the local scope, the analysis it will think it's dealing with `List` from the standard library.
+- There is no special support for functors. So with `Hashtbl.Make(...)` the builtin model will not apply. So the analysis won't report that the following can throw `Not_found`:
```
fhammerschmidt: functors --> module functions

fhammerschmidt: Hashtbl.Make is not even in the Stdlib. I am also not sure we should go into such detail here. Maybe that line and the example below can be omitted or at least simplified to one sentence?

pages/docs/manual/v12.0.0/editor-plugins.mdx
```diff
@@ -40,7 +40,89 @@ The code analysis provides extra checks for your ReScript project, such as detec
...
+#### Limitations
+
+- The libraries currently modeled are limited to `Array`, `Buffer`, `Bytes`, `Char`, `Filename`, `Hashtbl`, `List`, `Pervasives`, `Str`, `String` from the standard library, and `bs-json`, and `Json` from `Js`. Models are currently vendored in the analysis, and are easy to add (see [`src/ExnLib.ml`](src/ExnLib.ml))
+- Generic exceptions are not understood by the analysis. For example `exn` is not recognized below (only concrete exceptions are):
+
+```rescript
+try (foo()) { | exn => throw(exn) }
+```
+
+- Uses of e.g. `List.hd` are interpreted as belonging to the standard library. If you re-define `List` in the local scope, the analysis it will think it's dealing with `List` from the standard library.
```
fhammerschmidt: `List.head` as it is called now.

fhammerschmidt (changes requested):

fhammerschmidt (approved):

PR #1125, opened 2025-10-31
Initial blogpost on Rewatch

nojaf:
Supersedes https://github.com/rescript-lang/rescript-lang.org/pull/1124

### DIFF
--- a/_blogposts/2025-11-04-reforging-build-system.mdx
+++ b/_blogposts/2025-11-04-reforging-build-system.mdx
@@ -1,6 +1,6 @@
 ---
 author: rescript-team
-date: "2025-11-11"
+date: "2025-11-04"
 previewImg: /static/blog/rescript-12-reforging-build-system.png
 badge: roadmap
 title: "Reforging the ReScript Build System"


_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,266 @@
...
+### Slower Builds as Projects Grew
+
+As codebases grew, especially in monorepo setups with multiple packages, the old build system struggled to keep up. The module resolution had to search through many directories, and the dependency tracking wasn't sophisticated enough to avoid unnecessary work.
+
+These weren't just minor inconveniences. They were real productivity drains that got worse as projects scaled.
+
+## The Intelligence Behind ReWatch
+
+ReWatch takes a fundamentally different approach to building your code. At its core is a sophisticated understanding of what actually needs to be rebuilt when files change. Let's break down how it works.
+
```
zth: I really like that we're going as far in as `cmi`. It's great with deeper engineering content.

One thing that struck me is that if we're going this deep, it might be worth adding 1-3 lines to this section about how a ReScript program is compiled, at a high level. Mostly because it can be bit different to other languages. Might be worth mentioning one or two of these things:
- All ReScript files are compiled one-by-one, in isolation, and produce individual artifacts, just for that file. There's no shared compiler build state across modules.
- The module graph is a DAG - dependency cycles aren't allowed (which is a big factor in how compilation can be efficient)

Might be too much to include both, and you might not agree it's needed. But a thought at least.

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,266 @@
...
+### Faster Module Resolution with Flat Directory Layout
+
+ReWatch employs another clever optimization for module resolution. When building your project, it copies all source files to a flat directory structure at `lib/ocaml/`. Instead of maintaining the original nested directory structure, every module ends up in one place.
+
+Think of it like organizing a library. The old approach was like having books scattered across multiple rooms and floors. To find a specific book, you'd need to check each room. ReWatch's approach is like putting all the books in one room. Finding what you need is instant.
+
+**Why this matters:**
+
+- Module lookup becomes a single directory operation
+- The filesystem cache is more effective when files are adjacent
```
zth: Just because _someone_ might comment on it - file systems struggle when there are _too many_ files in a folder as well. But, since Rewatch takes ownership over where the files end up, sharding the layout later on as a perf optimization can be done entirely by Rewatch, and transparently to the user.

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,266 @@
...
+let defaultSize: size
+```
+
+With an explicit interface file, the `.cmi` is generated from the `.resi` file. Any changes to `Button.res` that don't affect the `.resi` contract will never trigger dependent recompilations. This gives you maximum build performance for frequently-changed modules.
+
+**Bonus for React developers:** Using `.resi` files for your React components has another benefit. During development, when you modify the component's internal implementation without changing the interface, React's [Fast Refresh](https://www.gatsbyjs.com/docs/reference/local-development/fast-refresh/) can preserve component state more reliably. Combined with ReWatch's intelligent rebuilding, this makes for an exceptionally smooth development experience.
+
+### Faster Module Resolution with Flat Directory Layout
+
+ReWatch employs another clever optimization for module resolution. When building your project, it copies all source files to a flat directory structure at `lib/ocaml/`. Instead of maintaining the original nested directory structure, every module ends up in one place.
```
zth: Very much a nit, but might be worth not writing out `lib/ocaml/`. Just because seeing `ocaml` might be confusing in a way that's not relevant to the article (why is that folder called ocaml? is it ocaml?).

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,266 @@
...
+
+ReWatch was designed from the ground up with monorepos in mind. File watching works correctly across all packages, detecting changes wherever they occur. The formatter is also properly integrated, working seamlessly across the entire monorepo structure.
+
+These aren't just nice-to-haves. For teams working with multiple packages, proper monorepo support means the build system finally works the way you'd expect.
+
+## What This Means for Your Workflow
+
+The techniques described above work together to create a build system that feels fundamentally different in practice. The most common operations during development (refactoring internal logic, fixing bugs, tweaking implementations) typically result in sub-second rebuilds. Less common operations that truly do require cascade recompilations (API changes, type modifications) benefit from parallel compilation and better module resolution.
+
+There's a threshold in human perception where interactions that complete in under 2 seconds feel instant. ReWatch aims to keep most of your development builds under this threshold, keeping you in flow rather than waiting for compilation.
```
zth: Is 2 seconds really true? IIRC 100ms is the low threshold for what a human can even detect, and after 1s it's generally noticable. But I might be wrong here.

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,266 @@
...
+```toml
+[install]
+linker = "hoisted"
+```
+
+We're continuing to test compatibility across different environments and configurations. If you encounter issues with any package manager, please report them so we can address them.
+
+## Using the Legacy Build System
+
+For projects that need it, the legacy bsb build system remains available through the `rescript-legacy` command. This is a separate binary, not a subcommand.
```
zth: Maybe add "remains available throughout the v12 release cycle" or something like that? Although I guess we haven't officially decided on the time frame yet.

zth (approved):
This is amazing! I love the technical depth of the article. We should really aim for more of these technically deeper articles.

I left a few comments/suggestions, do with it what you will.

_blogposts/2025-11-11-introducing-rewatch.mdx
cknitt: I would expect that we remove it in v13. This will help us get rid of so much legacy cruft.

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,267 @@
+---
+author: rescript-team
+date: "2025-11-11"
+badge: roadmap
+title: "ReWatch: A Smarter Build System for ReScript"
```
cknitt: So far, it was always called "Rewatch", not "ReWatch" AFAIK.

In general, I would like to deemphasize that name as it is (as you mention below) the internal codename for the new build system and CLI and is not user-visible anywhere.

So maybe we can omit it from the title altogether?

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,267 @@
...
+// Button.res
+type size = Small | Medium | Large
+
+let make = (~size: size, ~onClick) => {
+  // component implementation
+}
+
+let defaultSize = Medium
+
+%%private(let internalHelper = () => {
```
cknitt: `%%private` in its current state is quite weird, and I would prefer to avoid it in examples when possible. Maybe we can just have a .resi instead, which I would consider best practice?

cknitt (commented):
Excellent technical deep dive!

Maybe worth mentioning:

- What speedup can be expected in practice compared to bsb when building a project from scratch, not incrementally? From our projects (fairly large monorepos), as far as I recall, about a factor of 2.
- It's not just a new build system, but also a completely new CLI with a much better / more modern user experience than before (subcommands / help).

_blogposts/2025-11-11-introducing-rewatch.mdx
```diff
@@ -0,0 +1,267 @@
...
+**Why this matters:**
+
+- Module lookup becomes a single directory operation
+- The filesystem cache is more effective when files are adjacent
+- Cross-package references are as fast as local references
+- The compiler spends less time searching and more time compiling
+
+The small cost of copying files upfront is paid back many times over through faster compilation.
+
+### Wave-Based Parallel Compilation
```
cknitt: That sounds quite cool BTW! 🙂

fhammerschmidt (approved):
Reads extremely nice now!

PR #1127, opened 2025-11-05
perf: convert blog article image from png to webp

jderochervlk:
I noticed the latest blog post image was around 1.5mb. Switching it to webp brings it down just over 200KB. 

I'm planning on getting Vite setup to automatically handle images for us.


fhammerschmidt:
Yeah converting them in the build would be nice, because outside of the browser, webp support is still not great.

PR #1128, opened 2025-11-06
Regex and dict native syntax + type spread pattern match blog post

### LINKED ISSUE #1088: another syntax post about

tsnobip:
fixes #1088

### DIFF
--- a/_blogposts/2025-11-18-regex-dicts-variant-type-spread.mdx
+++ b/_blogposts/2025-11-18-regex-dicts-variant-type-spread.mdx
@@ -1,0 +1,94 @@
+---
+author: rescript-team
+date: "2025-11-18"
+badge: roadmap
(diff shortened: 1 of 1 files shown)


fhammerschmidt:
It's good but a tad bit short. Maybe we should merge it with the nested record types post?

tsnobip:
> It's good but a tad bit short. Maybe we should merge it with the nested record types post?

agreed.

I mean we could also just write more on the topic, but I couldn't find anything informative to add. Don't hesitate if you find anything I could write about @fhammerschmidt!

fhammerschmidt:
Ok, I merged the issue description.
@cknitt told me he were not able to start this yet anyways. So would be cool if you could overtake.

jderochervlk:
Are we still planning on publishing this? I think we should.

PR #1130, opened 2025-11-07
Rough outline of general v12 blog post

nojaf:


fhammerschmidt:
This is quite good already. Not sure if we can do without all those OCaml mentions. But I think it's not relevant to the user on whatever OCaml version we build the compiler.

Also not sure about the sponsorship stuff. Earlier this year I wanted to implement a Stripe form but somehow never could verify correctly with them as a representative of the ReScript Association. Maybe we should just drop the email: `contact \<at> rescript-association.org`

Also: you want to release on the 17th already?

nojaf:
> Also not sure about the sponsorship stuff. 

I don't think we need a grand plan first to mention that people can get in touch with us.
Whether we have some tool or system in place is irrelevant. 

> Also: you want to release on the 17th already?

Not really, this is just a date in the future for now.

markdown-pages/blog/release-12-0-0.mdx
```diff
@@ -0,0 +1,221 @@
...
+They continue to work in v12 with warnings but will be removed in v13.
+ReScript now prefers unified operator syntax: `&&&`, `|||`, `^^^`, and `~~~` cover AND, OR, XOR, and NOT for both `int` and `bigint`.
+
+### API naming alignment
+
+APIs that previously ended with `Exn` now end with `OrThrow`.
+Examples include `Option.getOrThrow`, `List.headOrThrow`, `BigInt.fromStringOrThrow`, and `JSON.parseOrThrow`.
+The change clarifies that these functions throw JavaScript errors, aligning the naming with the language’s semantics.
+The deprecated `*Exn` variants remain available in v12 to ease the transition, and the codemod bundled with the migration tool can perform a mechanical rename.
+Be aware that `Result.getOrThrow` now throws a JavaScript `Error`, so update any exception handling logic that depended on OCaml exception names.
```
tsnobip: while we're there, we should likely also mention the rework of JS exceptions from `Exn.Error(Error.t)` to `JsExn(unknown)` to better model JS exception payloads that can be of any type and potentially not Errors.

nojaf: @tsnobip as you are like the architecture of this devilry, would you mind tossing in a line here?

tsnobip: @nojaf sorry I missed that notification.
``````suggestion
Note that `Result.getOrThrow` now throws a JavaScript exception, so please update any exception-handling logic that relied on OCaml exception names. We also revamped the API around JS exceptions and errors.

We’ve renamed JS errors to `JsError.t` to better distinguish them from the `Result.Error` variant. Since JavaScript allows throwing anything (not only proper Error objects), the previous way of catching JS exceptions was unsafe:
```res
let foo = switch myUnsafeJsResult() {
| exception Exn.Error(e) => Error(e->Error.message)
  // ☝️ this will crash if `e` is undefined or null,
  // or return undefined if `e` has no `message` property
| myRes => Ok(myRes)
}
```
Additionally, the coexistence of `Result.Error` and the `Error` module caused confusion.

The new recommended pattern is:
```res
let foo = switch myUnsafeJsResult() {
| exception JsExn(e) => Error(e->JsExn.message->Option.getOr("[no message]"))
| myRes => Ok(myRes)
}
```
Utility helpers for working with JS errors are now in `JsError`, eg:
```res
JsError.throw(JsError.RangeError.make("the payload should be below 16"))
```
``````
What do you think?

Comment (nojaf) [reply markdown-pages/blog/release-12-0-0.mdx]: Works for me: "`JsError.t` to better distinguish them from the `Result.Error`" is a good line.

File: markdown-pages/blog/release-12-0-0.mdx [hunk 2]
```diff
@@ -0,0 +1,253 @@
...
+}
+```
+Additionally, the coexistence of `Result.Error` and the `Error` module caused confusion.
+
+The new recommended pattern is:
+```res
+let foo = switch myUnsafeJsResult() {
+| exception JsExn(e) => Error(e->JsExn.message->Option.getOr("[no message]"))
+| myRes => Ok(myRes)
+}
```
tsnobip, line 71: @nojaf I'm really sorry I forgot `Error.message` already returns an option, so the example can be simplified a bit:
``````suggestion
let foo = switch myUnsafeJsResult() {
| exception Exn.Error(e) => Error(e->Error.message)
  // ☝️ this will crash if `e` is undefined or null
| myRes => Ok(myRes)
}
```
Additionally, the coexistence of `Result.Error` and the `Error` module caused confusion.

The new recommended pattern is:
```res
let foo = switch myUnsafeJsResult() {
| exception JsExn(e) => Error(e->JsExn.message))
// ☝️ this is now safe even `e` is undefined or null
| myRes => Ok(myRes)
}
``````

fhammerschmidt:
superseded by #1144

Issue #1121, opened 2025-10-17
Document new operators


fhammerschmidt:
- [ ] `**` for [exponentiation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation), supports `float`, `int` and `bigint`
- [ ] `%` for [remainder](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Remainder) (in fact not modulo, see: https://bigmachine.io/career/mod-and-remainder-are-not-the-same), supports `float`, `int` and `bigint`
- [ ] `~~~` for [bitwise NOT](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NOT), supports `int` and `bigint`
- [ ] `|||` for [bitwise OR](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR), supports `int` and `bigint`
- [ ] `&&&` for [bitwise AND](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND), supports `int` and `bigint`
- [ ] `^^^` for [bitwise XOR](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR), supports `int` and `bigint`
- [ ] `<<` for [left shift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift), supports `int` and `bigint`
- [ ] `>>` for [right shift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Right_shift), supports `int` and `bigint`
- [ ] `>>>` for [unsigned right shift](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unsigned_right_shift), supports `int` only

Both add them to the [syntax lookup](https://rescript-lang.org/syntax-lookup) and to their corresponding section (float, int, bigint) in https://rescript-lang.org/docs/manual/v12.0.0/primitive-types.

Issue #1050, opened 2025-06-13
Document %ffi


fhammerschmidt:
- [ ] Add it to syntax lookup
- [ ] Adapt https://rescript-lang.org/docs/manual/v12.0.0/embed-raw-javascript to use %ffi instead of %raw

fhammerschmidt:
The only difference between `%raw` and `%ffi` is that with `%ffi` the arity is checked, so it is safer and should be preferred over raw.

See error in [playground](https://rescript-lang.org/try?version=v12.0.0-alpha.13&module=esmodule&code=DYUwLgBGIM5gSgQwO4C4IEsB2kC8A+THCXCAUgCcUAKAAwDMBXLAYzAwHssJEATX6gA8ANBACeuAKwBKAN4QK4RhW6CIAanEQAvrWkAofaEjQ4AMXoZ02PIRsly9S3Sat2XHvyGiJM+YrBlVQ0tXQMgA).

PR #1135, opened 2025-11-21
docs: remove deprecated content

jderochervlk:


markdown-pages/docs/manual/typescript-integration.mdx
```diff
@@ -264,19 +264,6 @@ export type inputFocusEvent = React.FocusEvent<HTMLInputElement>;
 
 More complete example shims can be found [here](https://github.com/rescript-lang/rescript-compiler/blob/master/jscomp/gentype_tests/typescript-react-example/src/shims/).
 
-## Deprecated features
```
fhammerschmidt, line 267: I am afraid we need to keep those still, right @cometkim ?

fhammerschmidt:
Syntax Lookup:
- [x] deprecate `%re` (we have native regex syntax now!)
- [x] deprecate `+.`, `-.`,`*.`, `/.`
- [x] change `mod` to `%`
- [x] deprecate `%raw` in favor of `%ffi` which checks the arity of the JavaScript function.
- [x] remove the whole Built-In Functions section, we have separate ignore functions which are documented in the API docs

there is a separate issue for all the new operators: #1121 
and another one for `%ffi`: #1050

markdown-pages/docs/manual/typescript-integration.mdx
jderochervlk: I set this back to normal.

PR #1142, opened 2025-11-25
Use yarn

cknitt:
After installing with npm, with the current lockfile, I was getting

> Error: Cannot find module @rollup/rollup-darwin-arm64. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.

I think we should move to yarn anyway to match what we have in the main `rescript` repo.

This also helps us to get rid of patch-package by using yarn's built-in patching functionality.


tsnobip (approved):
LGTM! thanks a lot @cknitt!

PR #1158, opened 2025-12-01
Enable JSX preserve mode

jderochervlk:
Turning on JSX preserve mode improves HMR and sets us up to add the React Compiler.

We have a few scripts that are written in ReScript that are run using Node. These files need to be renamed from `.jsx` to `.mjs` before they can be run and bundled to fix the rescript runtime having a different file extension.


cristianoc:
Oh this works now? Nice!!

fhammerschmidt (approved):

PR #1159, opened 2025-12-02
ci: Deploy to Cloudflare using a Github Action

jderochervlk:
This PR switches builds over from being configured in the Cloudfront dashboard to a Github action, which means it's now Infrastructure as Code. We can now also use a Github action to manually fire off a build, which is nice. Having this happen on a Github action will give us access to the deployment URL, which will be needed if we want to setup Playwright or Chromatic for automated testing (which would be great!).

This also tested out an example Cloudflare function, which you can see at https://d0b34a60.rescript-lang.pages.dev/echo?foo=bar. This doesn't mean that the app is using SSR, but Cloudflare pages can still have access to functions, which I plan on using to create dynamic open graph images. The function has been removed after testing. I wanted to make sure that Cloudflare could handle `.jsx` files.

This also changes some of the commands used for builds and development to simplify things. Now you can run `yarn dev` to kick off concurrent builds for ReScript, Vite, and Wrangler (which will be used to develop functions locally).

The existing Cloudflare job is expected to fail and will need to be removed alongside of merging in this PR (probably right before).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,8 +10,6 @@ on:
         required: true
         default: "preview"
   pull_request:
-    branches:
-      - master
 
 jobs:
   deploy:


cometkim (commented):
Deploying via actions is good.

It would be helpful to [migrate the pages project to a workers project](https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/) immediately afterward.

The [`@cloudflare/pages-plugin-vercel-og`](https://developers.cloudflare.com/pages/functions/plugins/vercel-og/) project provides excellent support for dynamic OG using a combination of [Satori](https://github.com/vercel/satori) and [HTMLRewriter](https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/). Also, a change to the CDN cache configuration is required to reduce unnecessary function calls.

jderochervlk:
> Deploying via actions is good.
> 
> It would be helpful to [migrate the pages project to a workers project](https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/) immediately afterward.
> 
> The [`@cloudflare/pages-plugin-vercel-og`](https://developers.cloudflare.com/pages/functions/plugins/vercel-og/) project provides excellent support for dynamic OG using a combination of [Satori](https://github.com/vercel/satori) and [HTMLRewriter](https://developers.cloudflare.com/workers/runtime-apis/html-rewriter/). Also, a change to the CDN cache configuration is required to reduce unnecessary function calls.

Why would we want to use workers? Wouldn't pages be best for the static content?

PR #1179, opened 2026-01-30
Fix guideline link

### LINKED ISSUE #1178: Link Broken - GUIDELINES  Publishing ReScript npm packages
the`Publishing ReScript npm packages` link on the page https://rescript-lang.org/packages?search= is broken.

https://rescript-lang.org/docs/guidelines/publishing-packages is an empty page.

lungarella-raffaele:
Fixes #1178

Added `docs/guidelines` to the rendering conditional so the page uses `DocsLayout` instead of falling through to `React.null`

The page has still a strange artifact for the "Back to packages" link, what would be the best fix?

Screenshot 2026-01-30 at 14 58 16

### DIFF
--- a/app/routes/MdxRoute.res
+++ b/app/routes/MdxRoute.res
@@ -299,7 +299,8 @@ let default = () => {
       </>
     } else if (
       (pathname :> string)->String.includes("docs/manual") ||
-        (pathname :> string)->String.includes("docs/react")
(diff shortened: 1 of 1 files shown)


fhammerschmidt:
ah deplyoment would need rebase, nevermind

fhammerschmidt (approved):

Issue #1162, opened 2025-12-19
navigation issues


chiang03:
### Example

https://rescript-lang.org/docs/manual/api/stdlib/promise#value-catch

### Reached by

clicking on links on the sidebar. Doing so on `API` pages adds hash to the URL, but doesn't on `Manual` or `React` pages

### Problems (already fixed)

- already fixed
  - `API` pages are sometimes not scrollable. It seems to be cycling between scrollable and not on each click on links on sidebar.

### Example

https://rescript-lang.org/docs/manual/api/stdlib/promise/#value-catch

### Reached by

1.  directly input (e.g. copy-paste) the URL of either case into the browser URL box, or
2.  refresh page

### Problems

1. on `Manual` and `React` pages, there are no buttons to previous/next chapter at the bottom
2. sidebar doesn't highlight the current page on any Manual, React, or API page.

---

There are error messages in browser dev tools console:

- `Minified React error #418` https://react.dev/errors/418?args[]=text&args[]=

  - logged on initial load of any page, such message does not appear when navigating via the sidebar

- `GET https://rescript-lang.org/docs/manual/introduction.data net::ERR_ABORTED 503 (Service Unavailable)`
  - logged every time any link on sidebar is hovered
  - only occurs on `Manual` and `React` pages, not on `API` pages
  - error URL depends on which link is hovered

chiang03:
The problem that the page being not scrollable is fixed, all other problems persist.

PR #1180, opened 2026-01-30
Fix api doc page not scrollable

lungarella-raffaele:
Fixes part of #1162
This addresses a sidebar state management issue where the scrollbar behavior and sidebar state aren't properly coordinated.

I **think** the intended behavior should be:

- Desktop/tablet: sidebar always open, page scrollbar visible
- Mobile: sidebar toggles open/closed, scrollbar locked when sidebar is open

However, the current implementation has some issues:

1. On page render, `isSidebarOpen` flashes from `true` to `false` even when the sidebar is clearly visible
2. Clicking the sidebar toggle affects both `scrollLock` and `isSidebarOpen` on all screen sizes, even though this behavior should only apply to mobile (I think)
3. The sidebar state issues aren't immediately obvious because CSS media queries independently control sidebar visibility, hiding the underlying state management problems

I've updated the logic so that sidebar state management primarily matters on mobile, with the sidebar defaulting to open on larger screens, but I had to add an hook to check if we are on mobile with an hardcoded breakpoint

This makes the docs more usable, but I'd love feedback on how to improve it! 🙂

### DIFF
--- a/src/ApiDocs.res
+++ b/src/ApiDocs.res
@@ -288,13 +288,19 @@ module DocstringsStylize = {
 
 let make = (props: props) => {
   let (_, setScrollLock) = ScrollLockContext.useScrollLock()
-  let (isSidebarOpen, setSidebarOpen) = React.useState(_ => true)
+  let isMobile = Hooks.useIsMobile()
+  let (isSidebarOpen, setSidebarOpen) = React.useState(_ => false)
 
-  let toggleSidebar = () =>
-    setSidebarOpen(prev => {
-      setScrollLock(_ => !prev)
-      !prev
-    })
+  React.useEffect(() => {
+    setSidebarOpen(_ => !isMobile)
+    None
+  }, [isMobile])
+
+  let toggleSidebar = () => {
+    let nextSidebarOpen = !(isSidebarOpen && isMobile)
+    setSidebarOpen(_ => nextSidebarOpen)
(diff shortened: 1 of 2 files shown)


jderochervlk:
I think some of this, or all of this, is fixed by: https://github.com/rescript-lang/rescript-lang.org/pull/1156

@fhammerschmidt mentioned another issue, but I think it's best to try and get that merged in before any other scrolling or layout fixes.

fhammerschmidt:
Yeah but only if you dont introduce new layout issues check my last comment)

jderochervlk:
I closed out the other PR, so I think we should go ahead with this fix. @lungarella-raffaele if you pull from master I think we should be able to have a cloudflare preview run so we can test things.

raffaele-lungarella:
Let me know if I can do something to fix the build :)

jderochervlk:
> Let me know if I can do something to fix the build :)
> 

It's a permission issue on the GitHub action and some missing env variables that aren't accessible by forks. I feel like we should be able to figure it out.

fhammerschmidt:
I believe #1186 solves this as well?

Bushuo:
@lungarella-raffaele Sorry, I didn't notice that there already were PRs open to address this issue...

raffaele-lungarella:
No worries, I will close the pr since the bug is fixed 🎉

PR #1186, opened 2026-02-06
fix: scroll lock on link click

Bushuo:
This fixes the scroll lock when clicking on sidebar links in the right sidebar.

The problem is, we only want scroll-lock on mobile viewports, because the sidebar (dropdown nav menu) overlays the screen there. On other viewport sizes the sidebar is in a different location and always open.

Because the sidebar component is heavily reused I could not figure out a cleaner solution without media queries.

### DIFF
--- a/src/ApiDocs.res
+++ b/src/ApiDocs.res
@@ -46,7 +46,7 @@ type item =
 
 module RightSidebar = {
   @react.component
-  let make = (~items: array<item>, ~onClick) => {
+  let make = (~items: array<item>, ~onLinkClick=?) => {
     items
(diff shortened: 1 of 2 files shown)


jderochervlk:
> Because the sidebar component is heavily reused I could not figure out a cleaner solution without media queries.

I've been running into this issue. It's hard to move things around or add to the navigation right now because of this. I tried to do a refactor in place, but it was too difficult to get it correct for each template. 

This fix is good for now, and I'm planning on breaking up the sidebar layouts into more focused templates and reusable components.

PR #1187, opened 2026-02-06
Deploy PR from fork

fhammerschmidt:
This enables deployments of PRs from external contributors (on forks) if and only if a maintainer adds the label
"ok-to-deploy". I believe this is the best balance between security and convenience. Because adding more and more people to the core maintainers to have a deployment is even worse IMO.


fhammerschmidt:
Closing in favor of #1188

PR #1191, opened 2026-02-07
fix(open graph images): added full url, file extension, and switch from query params to paths

jderochervlk:
It turns out that og images are finicky. After testing on several platforms I realized my first attempt wasn't correct.

- OG images need to have a file extension
- Some platforms (bluesky and text messages) ignore query params
- The base URL needs to be included on some platforms


jderochervlk:
Tested on what's app

jderochervlk:
Bluesky

jderochervlk:
Text message.

jderochervlk:
Tested on X

fhammerschmidt (approved):

PR #1199, opened 2026-02-21
Add docs for @notUndefined decorator and its usage

Bushuo:
This PR adds docs for the `@notUndefined` decorator.


fhammerschmidt (approved):
Thanks!

PR #1202, opened 2026-02-26
Fix path normalization in API docs

### LINKED ISSUE #1162: navigation issues

fhammerschmidt:
Fixes #1162

### DIFF
--- a/src/ApiDocs.res
+++ b/src/ApiDocs.res
@@ -99,14 +99,15 @@ module SidebarTree = {
     open ReactRouter
 
     let location = useLocation()
+    let currentPathname = (location.pathname :> string)->Url.normalizePath
 
     let moduleRoute = `${apiDocsRootPath}/${(location.pathname :> string)
       ->String.replace(`/docs/manual/api/`, "")
       ->String.split("/")
       ->Array.get(0)
-      ->Option.getOr("stdlib")}`
+      ->Option.getOr("stdlib")}`->Url.normalizePath
 
-    let isCurrentlyAtRoot = (location.pathname :> string) == moduleRoute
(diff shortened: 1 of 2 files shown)


fhammerschmidt:
@chiang03 can you check if this fixes your issues?

Test-Deployment: https://95c3d116.rescript-lang.pages.dev/

I don't really understand the part with the next/previous buttons since they show up for me?

chiang03:
The problems don't occur on the test deployment link.
Thank you for the fix.

PR #1211, opened 2026-03-19
tests: add more unit tests to cover markdown related components

jderochervlk:
Adding more tests before doing any more refactoring or changes.


fhammerschmidt:
Nice! Hopefully this does not get too annoying when doing visual changes.

PR #1212, opened 2026-03-19
chore: add AGENTS.md file

jderochervlk:
I used this as a base and then asked Claude to expand it.

https://forum.rescript-lang.org/t/best-practice-developer-copilot-rules-for-react/6231/2?u=jderochervlk

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,5 @@
+# Claude Code Guidelines
+
+All project guidelines, coding conventions, and agent instructions are maintained in a single shared file.
+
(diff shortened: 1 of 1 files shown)


fhammerschmidt:
Cool but I think for Claude itself you still need a CLAUDE.md as well, right?

jderochervlk:
> Cool but I think for Claude itself you still need a CLAUDE.md as well, right?

Apparently yes. I don't use Claude, but from what I can see in the docs it doesn't know about the `AGENTS.md` file unless you reference it in a `CLAUDE.md` file. I added that in.

PR #1220, opened 2026-04-02
tests: Stabilize screenshot tests and clean up flaky infrastructure

jderochervlk:
Fixes several issues with screenshot tests producing inconsistent or incorrect results.

I was getting different results running the tests locally versus the runs on CI, so I made some config changes to lock in the browser size and viewport. There were also some other issues I addressed to produce better and more stable results.

Key changes:**

1. **Removed ~60 stale/flaky screenshot baselines** - The bulk of this PR deletes a large number of `.png` screenshot files from `__tests__/__screenshots__/` across many test suites (Button, CodeExample, Blog, DocsLayout, Footer, SyntaxLookup, SidebarLayout, etc.). These were presumably unreliable or no longer needed.

2. **Stabilized screenshot comparison config** (`vitest.config.mjs`) - Added explicit screenshot configuration:
   - Forces `deviceScaleFactor: 1` to prevent DPI-related differences across machines
   - Sets `scale: "css"` for screenshot options
   - Adds tolerance thresholds (`threshold: 0.2`, `allowedMismatchedPixelRatio: 0.05`) to reduce false negatives
   - Disables the browser UI (`ui: false`)

3. **Added test-specific CSS overrides** (`styles/test-overrides.css`) - A new stylesheet imported only in `vitest.setup.mjs` that:
   - Disables all CSS animations and transitions (zero duration/delay) to eliminate timing-based flakiness
   - Hides scrollbars (`scrollbar-width: none`) so scrollbar rendering differences across environments don't break screenshots

4. **Fixed NavbarSecondary tests** - Switched from `<BrowserRouter>` to `<MemoryRouter initialEntries=["/docs/manual/introduction"]>`, giving the tests a deterministic URL context rather than relying on the browser's actual location.

5. **Removed video screenshot assertion** - In `MarkdownComponents_.test.res`, the `toMatchScreenshot("markdown-video")` call was removed (video rendering is inherently non-deterministic for screenshots).

6. **Removed auto-commit of screenshots from CI** (`.github/workflows/pull-request.yml`) - The `git-auto-commit-action` step that automatically committed updated screenshots on PRs was removed, preventing confusing auto-pushed changes in PRs.

7. **Updated `AGENTS.md`** - Added a rule clarifying that `.jsx` files should never be edited directly since they're compiler-generated by ReScript.

**Overall:** This is a testing infrastructure cleanup focused on making screenshot tests reliable and deterministic across different environments (local dev, CI, headless vs headed browsers) by eliminating the most common sources of visual diff noise.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -179,8 +179,10 @@ let default: unit => React.element
   })
   ```
 
-- Components requiring React Router context must be wrapped in `<BrowserRouter>`.
-- Run tests with `yarn vitest`.
+- Components requiring React Router context must be wrapped in `<MemoryRouter>`.
+- Do not use `<BrowserRouter>` in tests.
+- Run tests with `yarn vitest --browser.headless --run`.
+- Do not update snapshots without asking for confirmation.
 
 ## MDX Content
 


.github/workflows/pull-request.yml
```diff
@@ -47,8 +47,3 @@ jobs:
         run: yarn playwright install --with-deps
       - name: Vitest
         run: yarn ci:test
-      - name: Commit and Push changes
```
jderochervlk, line 50: This was a patch to keep the files in sync across local runs and CI to make sure that the CI versions were always used so they would be the same, but now they are the same in headless mode as they are running on chrome so this isn't needed.

PR #1238, opened 2026-04-09
fix: Add .resi file for CommunityRoute.jsx

jderochervlk:
By omitting the `.resi` file from this route it causes in app navigation to that page to not work due to the extra exports that React Router is not expecting. I'm surprised that it builds with the pre-render but the actual routing doesn't work after the build.


fhammerschmidt:
Ouch. We really need that private-by-default.

fhammerschmidt (approved):

jderochervlk:
> Ouch. We really need that private-by-default.

And I really need to get some e2e tests added!

PR #1246, opened 2026-04-10
docs: add page about opaque types

### LINKED ISSUE #543: Explain opaque types
Hi,

I often trend to link this section (https://reasonml.github.io/docs/en/type#opaque-types) from ReasonML to explain opaque types, but isn't ReScript and is more confusing that clarifying (mentioning type modules).

An explanation of the possibility to create a `type alias` without a specific type and keep it hidden should be more than enough.

jderochervlk:
Fixes https://github.com/rescript-lang/rescript-lang.org/issues/543 and https://github.com/rescript-lang/rescript-lang.org/issues/508

### DIFF
--- a/markdown-pages/docs/manual/opaque-types.mdx
+++ b/markdown-pages/docs/manual/opaque-types.mdx
@@ -1,0 +1,107 @@
+---
+title: "Opaque Types"
+description: "Opaque types syntax and examples for ReScript"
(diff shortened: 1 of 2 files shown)


jderochervlk:
I'm not sure we want to call them "opaque", "distinct", or "branded" types.

jderochervlk:
@codex

PR #1261, opened 2026-04-23
fix: fix contrast on playground settings text

jderochervlk:
## Before
image

image

## After
image

image


fhammerschmidt (approved):
Ah, so much better! Thanks!

PR #1282, opened 2026-04-28
Restore build configuration schema

fhammerschmidt:
this broke as well after the monorepo changes

Also I got a hydration error even after the globalThis fix so there is a workaround for it as well.

In the long run this needs to be axed.


jderochervlk (commented):
The Docson component is rough, but thanks for the fix! I do want to figure out a better solution to this,

jderochervlk (approved):

PR #1288, opened 2026-05-08
Restore missing branding page

fhammerschmidt:
also removed the BreadCrumbs.res file which was unused.
Brandcrumbs (c lowercase) is still there.


fhammerschmidt:
Thanks!

jderochervlk (approved):

PR #1323, opened 2026-07-27
Roadmap update for V13

fhammerschmidt:


apps/docs/markdown-pages/community/roadmap.mdx
```diff
@@ -8,16 +8,47 @@ order: 3
...
+
+The ReScript 13 language direction also includes:
+
+- dictionary spreads
+- record rest destructuring
+- inline record types in external definitions
+- first-class, type-safe tagged-template functions
+
+ReScript 13 also removes deprecated configuration, syntax, and standard-library APIs, defaults the module system to ES modules, and raises the minimum supported Node.js version to 22.
+
```
DZakh, line 49: I also want `JavaScript-Aligned Naming`:

- `bool` -> `boolean`
- `float` -> `number`
- `int` -> `int32`
- `Nullable` -> `Nullish`

I have issues with the agent misleading them without extra help, and I don't see much value in sticking to old names.

JonoPrest: I imagine it would cause a lot of migration friction though 🤔

fhammerschmidt: Never had such issues with GPT models.
I agree that an alignment makes sense for the users though.
I think this warrants a forum survey though. Or we have another way of aliasing everything.

DZakh: int32 already exists as I know. Most of the issues I had is with it using Nullable as T | null, while it's T | null | undefined.

I think an alias with deprecation message you can opt-out will be a good solution

jderochervlk (approved):

